PR #19, opened 2023-02-01
Ci/publish staging

hleonardtek:
The purpose is to deploy on staging whenever the main branch is pushed.
I put the deploy job in the publish workflow to avoid duplication.


fredericvilcot (approved):

.github/workflows/publish.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         with:
           images: ghcr.io/${{ github.repository }}
           tags: |
-            type=raw,enable=${{ endsWith(github.ref, github.event.repository.default_branch) }},value=nightly
+            type=raw,enable=${{ endsWith(github.ref, github.event.repository.default_branch) }},value=nightly-${{ github.sha }}
```
amimart, line 25: No please, do not touch the tags, we'll use the nightly.

.github/workflows/publish.yml
```diff
@@ -44,3 +44,27 @@ jobs:
...
+
+      - name: Set up kubernetes config
+        run: |
+          echo ${{ secrets.KUBECONFIG }} > config_enc
+          cat config_enc | base64 -d > config
+          rm config_enc
+
+      - name: Deploy staging
+        run: |
+          kubectl set image deployment/dataverse-portal dataverse-portal=${DATAVERSE_PORTAL_IMAGE}:nightly-${{ github.sha }} -n staging
```
amimart, line 70: We just need to update a metadata in order to trigger the rolling update:
```sh
kubectl patch deploy deploy dataverse-portal -p '{metadata:{"labels": {"commit": "${{ github.sha }}"}}}'
```

.github/workflows/publish.yml
```diff
@@ -44,3 +44,27 @@ jobs:
           push: true
           tags: ${{ steps.docker_metadata.outputs.tags }}
           labels: ${{ steps.docker_metadata.outputs.labels }}
+
+  deploy-k8s-staging:
+    runs-on: ubuntu-22.04
+    needs:
+      - publish-docker-images
```
amimart, line 51:
```suggestion
      - publish-docker-images
      if: github.ref == 'refs/heads/main'
```

We can condition the deploy to main pushes only.

amimart (changes requested):
Thanks, it works but I'll prefer avoid publishing new tags, we need to keep the nightly tag as is.

Instead, we only need to trigger the rolling update by updating a label on the Deployment.

.github/workflows/publish.yml
hleonardtek: Ok I just understood what you meant :open_mouth: .
Very smart thanks no need to change the tag  :100: !

ErikssonJoakim (approved):

.github/workflows/publish.yml
amimart: Yes there is 😆

amimart (approved):
Looks good! 👍

ccamel (approved):
Great! :+1:

hleonardtek:
@amimart @ccamel I forgot the namespace in the patch command.

ccamel (approved):

amimart (approved):

ccamel (approved):

PR #26, opened 2023-02-08
refactor: remove head elements & next images

fredericvilcot:
This PR removes `next` metadata in `Head` component.
It also removes useless images and code.

ℹ️ For now, metadata is declared as hard coded because we cannot use anymore runtime env variables in next13.


ccamel (approved):
👍

Issue #33, opened 2023-02-15
Add environment variables 


lolottetheclash:
### **Purpose**

The goal is to create a file containing the environment variables so that the application can access them and propagate them to all the pages of the site.

lolottetheclash:
This issue has been fixed by @ccamel .

PR #35, opened 2023-02-15
Feat: Add env variables

lolottetheclash:
👉 This PR implements this [issue](https://github.com/okp4/dataverse-portal/issues/33).

A *getConfig* helper function has been created to facilitate reusability.

The *server-only* package has been added to avoid the use of this helper on the client side; if it is used on the client side, then the developer will receive an error at build time.

⚠️ The error in the build is currently justified because the file containing the environment variables is not committed (the injection of these variables must be refined afterwards).


fredericvilcot:
Thanks for the work @lolottetheclash .
As we have moved our technical stack from `nextjs` to `vitejs`, I close this PR because it will not be used anymore.

Issue #34, opened 2023-02-15
Feature: Lateral bar


ccamel:
# Feature: Lateral bar

> **As a** user of the portal,
> **I want** to have a collapsible lateral bar on the left-hand side of the screen that contains navigation links and social network links, as well as a switcher for dark/light mode
> **So that** it allows me to quickly access important information and easily switch between different themes.

### Scenario: default display (desktop)

**Given that** I am a user of the web portal,
**And** I am using a *desktop* device,
**When** I access any page of the portal,
**Then** The lateral bar should be expanded on the left-hand side of the screen,
**And** should display navigation links to various sections of the portal (see mockup), social media links, and a button for sending feedback.
**And** should have the light theme selected.

### Scenario: collapse/expand

**Given that** I am a user of the web portal,
**When** I click the collapse icon on the expanded lateral bar,
**Then** the lateral bar should collapse and hide the navigation links and social network links, making more space available for the content area of the page (see mockup).
**When** I click the collapse icon of the collapsed lateral bar,
**Then** the lateral bar should expand on the left-hand side of the screen
**And** should display navigation links to various sections of the portal (see mockup), social media links, and a button for sending feedback.
**And** the lateral bar state (collapsed / expanded) is persisted in the browser for the next session.

### Scenario: theme switching

**Given that** I am a user of the web portal,
**When** I click the theme switcher button,
**Then** the theme of the entire portal should change to the alternate theme.
**And** the switcher button should visually reflect the currently selected theme.
**And** the theme is persisted in the browser for the next session.

### Scenario: links

**Given that** I am a user of the web portal,
**When** I click on the link: `Home`,
**Then** the portal should navigate to `/` and display a `Work In Progress` page
**When** I click on the link: `Explore`,
**Then** the portal should navigate to `/catalog` and display a `Work In Progress` page
**When** I click on the link: `My Data Spaces`,
**Then** the portal should navigate to `/dataspace` and display a `Work In Progress` page
**When** I click on the link: `Share Data & Services`,
**Then** the portal should navigate to `/sharing` and display a `Work In Progress` page
**When** I click on the link: `Create Knowledge`,
**Then** the portal should navigate to `/knowledge-builder` and display a `Work In Progress` page
**When** I click on the link: `Create Dataspace`,
**Then** the portal should navigate to `/dataspace-builder` and display a `Work In Progress` page
**When** I click on the link: `Build Applications`,
[...]

PR #36, opened 2023-02-15
Feat/lateral bar

fredericvilcot:
Implement #34 

👉 This PR adds a lateral bar on the left of the application. The content allows the user to navigate through the portal and to useful external links.

👉 It adds the theme selector that can move from light to dark (and vice versa). The theme is persisted into browser local memory.

👉 The bar is collapsible to gain more space into the page. (not persisted)

ℹ️ URLS are hardcoded for now, waiting for PR #35 to be implemented as env vars
ℹ️ This version of the lateral bar is only suitable for desktop


ccamel (approved):

src/components/switch/switch.tsx
```diff
@@ -0,0 +1,28 @@
+import type { FC } from 'react'
+import * as RSwitch from '@radix-ui/react-switch'
+import './switch.scss'
+
+type SwitchProps = {
+  isChecked: boolean
+  onCheckedChange: (checked: boolean) => void
+  icons: {
+    checked: JSX.Element
+    notChecked: JSX.Element
```
lolottetheclash, line 10: I think `unChecked` would be more appropriate, what do you think?

src/components/lateralBar/lateralBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+        className={classnames('okp4-dataverse-portal-lateral-bar-content-container', {
+          collapsed: !isExpanded
+        })}
+      >
+        {Object.entries(groupedRoutes).map((group, index) => (
+          <div
+            className={classnames('okp4-dataverse-portal-lateral-bar-content-block', {
+              collapsed: !isExpanded
+            })}
+            key={index}
```
lolottetheclash, line 107: Isn't it better to avoid the index as a key..?

src/components/lateralBar/lateralBar.tsx
```diff
@@ -0,0 +1,157 @@
...
+          isChecked={theme === 'light'}
+          onCheckedChange={handleCheck}
+        />
+      </div>
+      <div
+        className={classnames('okp4-dataverse-portal-lateral-bar-content-container', {
+          collapsed: !isExpanded
+        })}
+      >
+        {Object.entries(groupedRoutes).map((group, index) => (
```
lolottetheclash, line 102: I don't know what you think, but maybe the code could have been split into smaller JSX elements to improve readability? It's a detail ;)

src/components/buttons/feedback/feedbackButton.tsx
```diff
@@ -0,0 +1,13 @@
+import Image from 'next/image'
+import type { FC } from 'react'
+import feedbackIcon from '../../../../public/icons/feedback.svg'
+import './feedbackButton.scss'
+
+export const FeedbackButton: FC = () => (
```
lolottetheclash, line 6: Shouldn't we have avoided using FC as we discussed a few months ago? It is possible that this has changed in the meantime!

lolottetheclash (approved):
Great side bar👌 !!!
I gave you some feedback but it remains a personal opinion !

PR #40, opened 2023-02-17
build(docker): use index.html by default

amimart:
Update docker nginx configuration to retrieve the index.html on non existing resources.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -3,6 +3,10 @@ server {
     server_name  localhost;
 
     root /usr/share/nginx/html;
+
+    location /assets {
+        try_files $uri =404;
+    }
     location / {
         try_files $uri $uri/ /index.html;
     }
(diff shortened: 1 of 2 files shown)


nginx.conf
```diff
@@ -0,0 +1,9 @@
+server {
+    listen       80;
+    server_name  localhost;
+
+    root /usr/share/nginx/html;
+    location / {
+        try_files $uri /index.html;
```
ccamel, line 7:
```suggestion
        try_files $uri $uri/ /index.html;
```

ccamel: I think it's a best practice to serve static assets directly (with a specific route):

```
location /assets {
    # Serve static assets directly
    try_files $uri =404;
  }
```

ccamel (changes requested):

ccamel (approved):

fredericvilcot (approved):

PR #42, opened 2023-02-17
Feat/theme

fredericvilcot:
👉 This PR adds global app theming, palette and main style.

ℹ️ `theme` is persisted in `localStorage` thanks to `zustand` middleware.


ccamel (changes requested):

src/App.tsx
```diff
@@ -1,33 +1,18 @@
...
-  const [count, setCount] = useState(0)
-
-  const handleClick = useCallback(() => {
-    setCount(count => count + 1)
-  }, [])
-
+  const theme = useAppStore(store => store.theme)
+  const setTheme = useAppStore(store => store.setTheme)
+  const handleTheme = useCallback(() => {
+    setTheme(theme === 'light' ? 'dark' : 'light')
```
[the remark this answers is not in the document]
fredericvilcot: yes, @ccamel I agree with that point of view 😎

src/App.tsx
```diff
@@ -1,33 +1,18 @@
...
-
-  const handleClick = useCallback(() => {
-    setCount(count => count + 1)
-  }, [])
-
+  const theme = useAppStore(store => store.theme)
+  const switchTheme = useAppStore(store => store.switchTheme)
+  const handleTheme = useCallback(() => {
+    switchTheme()
+  }, [switchTheme])
```
ccamel, line 10: I'd remove this function. You could just pass `switchTheme` directly to the `onClick` handler (was the meaning of my previous comment) without any impact, since Zustand internally memoizes the resulting selector function, afaik.

fredericvilcot: Oops.. so true 😱

ccamel (approved):
Great! Thanks

PR #55, opened 2023-02-22
Build/environment variables

ccamel:
This PR introduces the parametrization of the HTML `favicon` `title` and `meta tags` (`keywords`, `description`, `version`...) using the ViteJS plugin [vite-plugin-html-config](https://github.com/ahwgs/vite-plugin-html-config). Obvisously, the values are injected in the HTML during build time.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -10,5 +10,14 @@ export default defineConfig({
     alias: {
       '@': path.resolve(__dirname, './src')
     }
-  }
+  },
+  build: {
+      rollupOptions: {
+        output: {
+          manualChunks: {
+            'react-vendor': ['react', 'react-dom', 'react-router-dom']
+          }
+        }
+      }
+    },
 })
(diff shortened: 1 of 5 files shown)


src/config.ts
```diff
@@ -0,0 +1,6 @@
+const config = {
+    apiUrl: "https://example.com/api",
+    appVersion: "1.0.0",
+  };
```
fredericvilcot, line 4: How do you handle those environment variables?
Will it be part of a next PR that will inject an `env.js` script in the `index.html` file?

vite.config.ts
```diff
@@ -2,13 +2,29 @@ import { defineConfig } from 'vite'
...
     alias: {
       '@': path.resolve(__dirname, './src')
     }
-  }
+  },
+  build: {
+      rollupOptions: {
+        output: {
+          manualChunks: {
+            'react-vendor': ['react', 'react-dom']
```
fredericvilcot, line 25: Maybe we could also add `react-router-dom` to this vendor, no?

ccamel: Ah ah, yes good catch. When I started the PR there was no router yet that's why. I add it of course.

fredericvilcot (approved):
Great, thanks @ccamel ❤️

PR #75, opened 2023-03-02
Feat/fp-ts

ccamel:
## Purpose

This PR resurfaces the initiative https://github.com/okp4/ui/issues/277, reinforcing the notion that adopting a functional paradigm could greatly benefit our application development in terms of brevity, expressiveness, code conciseness, maintainability, security, concurrency, and more.

To address this, this PR aims to introduce the functional library (https://github.com/gcanti/fp-ts to enable the expression of functional characteristics in our code. Additionally, it proposes implementing a `toEffectful` function that facilitates transitioning from the pure world to the impure world when using the domains built around [Zustand](https://github.com/pmndrs/zustand).

With this step, we are initiating a move towards a pure and side-effect-free approach to domain implementation. This approach involves firmly rejecting the use of exception throwing and async/await, and a resolutely functional coding style.

### **Notes**

- [Functional programming](https://en.wikipedia.org/wiki/Functional_programming)
- [gcanti/fp-ts](https://github.com/gcanti/fp-ts)
- [A gentle introduction to monads](https://kristofsl.medium.com/a-gentle-introduction-to-monads-bc583d41d95)

### DIFF
--- a/src/util/effect.ts
+++ b/src/util/effect.ts
@@ -4,18 +4,15 @@ import type { IO } from 'fp-ts/IO'
 // Defines the type of function that returns an IO or a Task also known as a computational function.
 export type ComputationalFunc = (...args: unknown[]) => IO<unknown> | Task<unknown>
 
+type EffectfulResultType<F extends ComputationalFunc> = ReturnType<F> extends IO<infer T>
+  ? T
+  : ReturnType<F> extends Task<infer U>
+  ? Promise<U>
+  : never
+
 // toEffectful is an utility function that converts a computational function (i.e a function that returns an IO or a Task) to a function that returns
 // its result directly (producing a side effect).
 export const toEffectful =
-  <F extends ComputationalFunc> (
-    fn: F
-  ): ((
-    ...args: Parameters<F>
-  ) => ReturnType<F> extends IO<infer T>
-    ? T
-    : ReturnType<F> extends Task<infer U>
-    ? Promise<U>
-    : never) =>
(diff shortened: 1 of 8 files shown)


ccamel:
Thanks @fredericvilcot for the improvements 😌

src/util/effect.ts
```diff
@@ -0,0 +1,21 @@
...
+  ): ((
+    ...args: Parameters<F>
+  ) => ReturnType<F> extends IO<infer T>
+    ? T
+    : ReturnType<F> extends Task<infer U>
+    ? Promise<U>
+    : never) =>
+  (...args: Parameters<F>) =>
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    fn(...args)() as any
```
fredericvilcot, line 21: I will have used `never` as return type because it represents the unreachable result of the first execution. 
By the way, it's more precise than `any` keyword.

fredericvilcot (commented):
Thanks @ccamel, this is the first step of the functional way ❤️
I just highlighted a minor remark, it's up to you 😌

src/util/effect.ts
ccamel: See 70ece341b0f447cf42b328c8e33fe45c7217b5fa. And tell me.

fredericvilcot: @ccamel 
Love it!
That's what I was meaning by the never usage: constraint type to narrowing conditionals until the never type.
Thanks

fredericvilcot (approved):

PR #76, opened 2023-03-03
Feat/environment parametrization

ccamel:
This PR updates the app to externalize certain parameters into an environment file, loaded as a JS script when the application is loaded. It's quite trivial. This will enable the easy substitution of values during deployment to customize the application to specific requirements.

### DIFF
--- a/public/assets/env.js
+++ b/public/assets/env.js
@@ -1,4 +1,4 @@
-const settings = {
+const APP_ENV = {
     urls: {
         'social:twitter': 'https://twitter.com/OKP4_Protocol',
         'social:linkedin': 'https://www.linkedin.com/company/okp4-open-knowledge-platform-for/',
--- a/public/assets/env.js
+++ b/public/assets/env.js
@@ -1,0 +1,10 @@
+const settings = {
+    urls: {
+        'social:twitter': 'https://twitter.com/OKP4_Protocol',
+        'social:linkedin': 'https://www.linkedin.com/company/okp4-open-knowledge-platform-for/',
+        'social:discord': 'https://discord.com/invite/okp4',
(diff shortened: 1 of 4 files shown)


amimart (approved):
Great, regarding the ability to update those values when deploying currently the only solution is by mounting a new version of this file on the filesystem, which is not ideal.

We could enhance the docker image with an entrypoint generating this file based on env vars, it could be easily done with `jq` but we can add this in a second time.

public/assets/env.js
```diff
@@ -0,0 +1,10 @@
+const APP_ENV = {
+    urls: {
+        'social:twitter': 'https://twitter.com/OKP4_Protocol',
+        'social:linkedin': 'https://www.linkedin.com/company/okp4-open-knowledge-platform-for/',
+        'social:discord': 'https://discord.com/invite/okp4',
+        'social:medium': 'https://blog.okp4.network',
+        'form:feedback': 'https://okp4.typeform.com/to/TNyFBH72',
+        'about:okp4': 'https://okp4.network',
+      }
+}
```
fredericvilcot, line 10: Just a little remark about record's keys syntax, I will have used nesting keys (e.g: `APP_ENV.urls.social.discord`) to allow destructuration usage instead of "bracket picking" that is less readable and manipulable.

fredericvilcot (approved):
Great 👍

public/assets/env.js
ccamel: Let's say I didn't see the need to introduce an additional layer of indirection (e.g., 'social') in the APP_ENV object. All URLs could be at the same level. 

However, you can still destructure the object if you prefer. To do this, you'll need to specify a new variable name for each URL. Alternatively, we could consider using property syntax to avoid this extra step.

PR #79, opened 2023-03-07
Feat/fp-ts part ][

ccamel:
This is a continuation of #75 that generalizes the functions to support objects containing computational functions. It is mainly intended to transform [Zustand](https://github.com/pmndrs/zustand) store using [fp-ts](https://gcanti.github.io/fp-ts/) into regular stores.


fredericvilcot (approved):
Great!

PR #90, opened 2023-03-10
Prevent env.js to be cached by browsers

ccamel:
Self-explanatory.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -4,9 +4,16 @@ server {
 
     root /usr/share/nginx/html;
 
+    location /assets/env.js {
+        add_header Cache-Control 'no-store';
+        add_header Cache-Control 'no-cache';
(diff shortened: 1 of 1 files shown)


fredericvilcot (approved):

amimart (commented):
Why this resource shouldn't be cached? We could apply a `must-revalidate` policy to ensure the proper version is used.

ccamel:
> Why this resource shouldn't be cached? We could apply a `must-revalidate` policy to ensure the proper version is used.

Good point. I don't really trust nginx's e-tag because it only uses the file's last modification date and size to make it, afaik. And since the file is pretty small, I don't think it'll mess with the performance too much...  But, okay, we can change it, if you want to.

amimart (approved):

PR #100, opened 2023-03-14
fix(sidebar): add condition to render alpha version

lolottetheclash:
This PR aims to hide the "ALPHA VERSION" mention on collapsed sidebar.
It should be visible only when the sidebar is expanded.

Currently we have:

Capture d’écran 2023-03-14 à 12 50 31

And it should be:

Capture d’écran 2023-03-14 à 12 50 49


ErikssonJoakim (approved):

ccamel (approved):
:+1:

fredericvilcot (approved):

PR #101, opened 2023-03-16
fix(card): remove card dimension restrictions

ErikssonJoakim:
This PR fixes a problem where in order to precise the card min-height and width you'd have to be higher in the css priority order


AnjaMis (approved):
Looks good to me!

lolottetheclash (approved):
Good for me !

PR #102, opened 2023-03-16
Feat/position homepage block header

ErikssonJoakim:
This PR changes how the header for each block on the home page is positioned. It also removes some redundant properties and adds header positioning for mobile and tablet breakpoints.


lolottetheclash (approved):
👍

Issue #82, opened 2023-03-07
Feature: Action cards home page


MdechampG:
# Feature: Action cards home page

> **As a** user of the portal authenticated or not,
> **I want** to see the 4 Action Cards on the home page but disabled
> **So that** I know the main actions I will be able to take to interact with OKP4 protocol when the related features will be deployed. 

### Scenario: Default display

**Given that** I am user of OKP4 portal,
**And** I am authenticated or not
**And** I am using a *desktop* device,
**When** I am on the home page,
**Then** I should see the Action cards 
**When** I hover the cards, I see the hover disable version (with Soon Available)
**And** the cards are not clickable

### Scenario: Mobile display

**Given that** I am user of OKP4 portal,
**And** I am authenticated or not
**And** I am using a *mobile* device,
**When** I am on the home page,
**Then** I should see the Action cards on the disabled version
**And** the cards are not clickable

### Scenario: Internationalization

**Given that** I am a user of the web portal on the catalog page,
**When** the selected language is English
**Then** texts should be displayed in English language,
`Share Data & Services`
`Create Knowledge`
`Create Data Spaces`
`Build applications`
`Soon available`

**Given that** I am a user of the web portal on the catalog page,
**When** the selected language is French
**Then** texts should be displayed in French language,
`Partager des Données et Services`
`Créer des Connaissances`
`Créer des Data Spaces`
`Construire des Applications`
`Bientôt disponible`

**Given that** I am a user of the web portal on the catalog page,
**When** the selected language is German
**Then** texts should be displayed in German language,
`Daten & Dienste teilen`
`Wissen schaffen`
`Data-Spaces erstellen`
`Anwendungen erstellen`
`Bald verfügbar`

### Mockups

Desktop

To complete

Mobile

To complete

Design system 

![image.png](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/046312b1-89f1-42f0-ad5c-ae97d7c1a4ed)

### Environment

- staging URL: https://app.staging.okp4.network/

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206541817) by [Unito](https://www.unito.io)

MdechampG:
Updated with faked data @AnjaMis

MdechampG:
See #122

MdechampG:
Test 24/03

Desktop : Firefox - Safari - Chrome
Mobile : Safari

Scenario: Default display : OK ✅

[...]

PR #113, opened 2023-03-20
Feat/add home page interact

AnjaMis:
This PR adds the "Interact" block on the homepage. There is an issue with the card background pictures and their scaling, this is a temporary solution approved by Marine. 

Link to the issue: https://github.com/okp4/dataverse-portal/issues/82

🚨The layout of the interact block is positioned higher than the layout and will be automatically placed in the right place when implementing the middle "Explore" block. Please disregard its location for now.


src/page/home/interact/interact.tsx
```diff
@@ -0,0 +1,75 @@
...
+import type { CardBackground } from '@/component/card/card'
+import { Card } from '@/component/card/card'
+import classNames from 'classnames'
+import './interact.scss'
+import '../i18n/index'
+
+type InteractCard = {
+  category: string
+  title: string
+  _tag: string
```
ccamel, line 13: Can be named `tag` instead of `_tag`, the underscore prefix referring to a meta-information, which is not the case here, as far as I understand.

```suggestion
  tag: string
```

ccamel (approved):
Good job! :+1:

src/style/theme.scss
```diff
@@ -22,6 +22,7 @@ $themes: (
     tertiary-button-background: $tertiary-button-background-light,
     tertiary-button-background-variant-1: $tertiary-button-background-light-variant-1,
     card: $card-light,
+    card-background: $card-background,
```
fredericvilcot, line 25: If this color is related to the disabled status, I will have added it more explicitly

```suggestion
    card-background-disabled: $card-background,
```

src/component/card/card.scss
```diff
@@ -1,10 +1,32 @@
...
     display: flex;
     align-items: center;
+
+    @each $background-key, $background-value in $backgrounds {
+      &.#{$background-key} {
+        background: themed($background-key)
+          url('@/asset/images/#{$background-value}.webp')
+          no-repeat;
+
+        &:hover {
```
fredericvilcot, line 22: I will have added a `cursor: pointer;` to highlight the idea that these cards are CTA

fredericvilcot (changes requested):
Great job! 👏
I have added some remarks to be taken in consideration because, IMHO, it improves a bit extensibility and readability 😌

src/style/theme.scss
lolottetheclash: I had a doubt about it yesterday

src/page/home/interact/interact.tsx
```diff
@@ -0,0 +1,64 @@
...
+      {interactCards.map(({ action, category, backgroundColor }, index) => (
+        <Card
+          background={backgroundColor}
+          key={index}
+          mainClassName="okp4-dataverse-portal-home-page-interact-card"
+        >
+          <div>
+            <h3>{action}</h3>
+            <h2>{category}</h2>
+            <div className="okp4-dataverse-portal-home-page-interact-card-disabled">
```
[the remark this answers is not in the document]
lolottetheclash: I understand what you mean but is it relevant to add the notion of `disabled` (which concerns the `Card` behavior itself) in `BackgroundCard` which only talks about the background colors? what do you think about it @fredericvilcot ?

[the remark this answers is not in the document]
fredericvilcot: Absolutely, I would say the `InteractCard` object. 😌

src/component/card/card.tsx
```diff
@@ -2,11 +2,23 @@ import type { FC } from 'react'
...
+export type CardBackground =
+  | 'primary-color'
+  | 'primary-color-variant-1'
+  | 'primary-color-variant-2'
+  | 'primary-color-variant-3'
+  | 'card'
+
 export type CardProps = {
   children: React.ReactElement
+  background?: CardBackground
```
ErikssonJoakim, line 14: For me this property is not needed. We can already add style to the card with the mainClassName property

ErikssonJoakim (changes requested):
Great work @AnjaMis ! :smiley: 👍  There are just some things I like you to look at. Let me know if you want more clarification

src/page/home/interact/interact.scss
```diff
@@ -0,0 +1,69 @@
+@import '@/style/fonts.scss';
+@import '@/style/mixins.scss';
+
+.okp4-dataverse-portal-home-page-interact-card {
+  text-align: center;
+  justify-content: center;
+  height: 155px;
```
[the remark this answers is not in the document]
AnjaMis: @lolottetheclash what do you think? Should I change it back?

src/component/card/card.tsx
AnjaMis: Hmm, for this one we decided to keep all the card background styles outside of interact.scss (basically keep everything in card.scss) as the background images are also stored there 🤔

src/page/home/interact/interact.scss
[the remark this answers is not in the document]
lolottetheclash: For the moment we stay like that, we will have to define if we allow the `Card` to grow if the content is too long or if we position an overflow or other.

src/page/home/interact/interact.tsx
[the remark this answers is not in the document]
AnjaMis: Done!

lolottetheclash (approved):

ErikssonJoakim (approved):
Good job 😺 🚀

src/style/palette.scss
```diff
@@ -40,6 +40,7 @@ $switch-light-variant-1: white;
 // cards
 $card-dark: #242433;
 $card-light: #f2f6f9;
+$card-background: #001526;
```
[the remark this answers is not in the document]
ErikssonJoakim: that works :)

src/page/home/interact/interact.tsx
```diff
@@ -0,0 +1,81 @@
...
+import type { CardBackground } from '@/component/card/card'
+import classNames from 'classnames'
+import { Card } from '@/component/card/card'
+import './interact.scss'
+import '../i18n/index'
+
+type InteractCard = {
+  action: string
+  category: string
+  disabled: boolean
```
fredericvilcot, line 13: `disabled` should be optional, so it will fe evaluated with `false` value  if undefined.

src/page/home/interact/interact.scss
```diff
@@ -0,0 +1,57 @@
...
+    color: white;
+  }
+
+  .okp4-dataverse-portal-home-page-interact-card-disabled-text {
+    display: none;
+    justify-content: center;
+  }
+
+  &:hover {
+    &.disabled {
```
fredericvilcot, line 39: With the conditional rendering usage, this could be removed.

src/component/card/card.scss
fredericvilcot: @AnjaMis 
Does this change has been taken into account?

fredericvilcot: it's ok I resolve the comment

fredericvilcot (changes requested):
@AnjaMis 
Please, can you add a transition on the hover so it will be smoother.
Thx

PR #114, opened 2023-03-21
fix: improve french i18n

lolottetheclash:
This PR fixes the french translation "Voir details" into "Voir Détails".


amimart (approved):
That's better 😉

PR #115, opened 2023-03-21
ci(dependabot): update assignees and reviewers

lolottetheclash:
Self explanatory.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,6 +34,9 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      - dependency-name: '*'
+        update-types: ['version-update:semver-major']
     reviewers:
       - ccamel
       - fredericvilcot
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,13 +5,15 @@ updates:
     schedule:
       interval: weekly
     reviewers:
-      - amimart
       - ccamel
       - fredericvilcot
       - ErikssonJoakim
+      - lolottetheclash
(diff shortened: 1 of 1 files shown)


amimart (approved):
🥳

.github/dependabot.yml
```diff
@@ -5,13 +5,14 @@ updates:
     schedule:
       interval: weekly
     reviewers:
-      - amimart
       - ccamel
       - fredericvilcot
       - ErikssonJoakim
+      - lolottetheclash
```
fredericvilcot, line 11: I will have added the whole team as reviewers to keep the consistency about updates knowledge 😌

ccamel (approved):

.github/dependabot.yml
```diff
@@ -21,9 +22,11 @@ updates:
...
     assignees:
       - ccamel
       - fredericvilcot
+      - lolottetheclash
     open-pull-requests-limit: 5
   - package-ecosystem: docker
```
fredericvilcot, line 31: If you don't mind, could please force `dependabot` configuration to only open PR about `docker` related to patches? Thanks 😎

fredericvilcot (changes requested):

.github/dependabot.yml
ccamel: @fredericvilcot I prefer to keep Dependabot PRs for all Docker images. This way, we can stay up-to-date with all available upgrades and security patches, especially regarding the `nginx` docker image we use.

ccamel: All right. Thanks.

Issue #117, opened 2023-03-21
Enhancement : App - fix sidebar & toolbar


ErikssonJoakim:
# Feature: Enhancement : App - fix sidebar & toolbar
> **As a**  desktop, tablet and mobile user of the Portal,
> **I want** to have easy access to the app sidebar & toolbar when navigating the application
> **So that** I don't have to scroll through long pages to access the sidebar & toolbar

### Scenario: all displays
**Given that** I am a desktop, tablet and mobile user of the Portal,,
**When** when I scroll through a page in any navigation item,
**Then** the sidebar & toolbar should be fixed when I scroll down / up
**So that** I always have access to menus at any time and anywhere in the app

## Requirements
This issue is easily testable on the Dataverse page, as it contains a large number of elements. 
We must be able to scroll the whole page to see all the cards, the size of the sidebar is the same as the screen without scroll inside, and the toolbar remains visible. 

### Environment

* staging URL: <https://app.staging.okp4.network/>

## Additional informations
None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206527953) by [Unito](https://www.unito.io)

fredericvilcot:
@MdechampG 
FYI

MdechampG:
Test 23/03

Desktop : Safari, Firefox, Chrome
Mobile : Safari
Tablet (with desktop) : Firefox

Scenario : all displays OK ✅ 

❌ When the screen is not high enough, we can't see the bottom of the bar as we can't scroll the lateral bar. 

<img width="1522" alt="image" src="https://user-images.githubusercontent.com/92780073/227256406-9b73057f-c4b8-4172-8f0b-0c3d9479173e.png">

fredericvilcot:
@ErikssonJoakim 
Please, could you take care of this issue?
Thx

ErikssonJoakim:
@fredericvilcot Yes I will look into it 🙂

MdechampG:
@ErikssonJoakim is it linked to the #132  ?

ErikssonJoakim:
@MdechampG No I've just merged a fix for the problem that you described above. #132 will be to test a part

MdechampG:
Thank you @ErikssonJoakim !

Test 28/03

Desktop : Safari, Firefox, Chrome
Mobile : Safari
Tablet (with desktop) : Firefox

Scenario : all displays OK ✅

PR #118, opened 2023-03-21
feat(layout): fix sidebar & toolbar position in viewport

ErikssonJoakim:
This PR makes the sidebar and toolbar have a fixed position.
- #117

By request I've also done the following:
- added a bottom padding 
- modified the sidebar nav block gaps 
- customized the scrollbar (transparency & size)

<details>
<summary>1440px</summary>

127 0 0 1_5173_dataverse (5)
127 0 0 1_5173_dataverse

</details>

<details>
<summary>768px</summary>

127 0 0 1_5173_dataverse (4)
127 0 0 1_5173_dataverse (1)

</details>

<details>
<summary>375px</summary>

127 0 0 1_5173_dataverse (3)
127 0 0 1_5173_dataverse (2)

</details>

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,8 +18,10 @@ const App: FC = () => {
             'expanded-sidebar': isSidebarExpanded
           })}
         >
-          <Toolbar />
-          <AppRouter />
+            <Toolbar />
+          <div className="okp4-dataverse-portal-page-wrapper">
+            <AppRouter />
+          </div>
         </div>
       </div>
     </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,7 +18,9 @@ const App: FC = () => {
             'expanded-sidebar': isSidebarExpanded
           })}
         >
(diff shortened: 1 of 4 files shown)


lolottetheclash (changes requested):
I think the `last-child` css selector is not robust: what will happen if there is not only the toolbar and the main page, but also other elements? the last child will not necessarily select the right element... don't you think @fredericvilcot ?

fredericvilcot:
> I think the `last-child` css selector is not robust: what will happen if there is not only the toolbar and the main page, but also other elements? the last child will not necessarily select the right element... don't you think @fredericvilcot ?

Yes, would be better to target the `className` IMHO

ErikssonJoakim:
Yes you are correct. I've added a wrapper class name to avoid potential issues. I've also reworked the layout padding to be able to position the scrollbar more to the right, please see updated images

ccamel (approved):
:+1:

ccamel (approved):

Issue #77, opened 2023-03-03
Home page catalog preview


MdechampG:
# Feature: Home page catalog preview ‘Explore’
> **As a** user of the portal on the home page
> **I want** to be able to see 4 resources of the dataverse 
> **So that** I see 4 interesting resources and it makes me want to discover more.

### Scenario: default display (desktop)
**Given that** I am a user of the web portal,
**When** I am on the home page (authenticated or unauthenticated)
**Then** I should see 4 cards of dataverse’s resources (see mock up) 
**And** I should see a Call To Action button “Explore the Catalog” 

### Scenario: See Details
**Given that** I am a user of the web portal on the home page,
**Then** I see the button "See Details" that are disabled

### Scenario: Explore the catalog
**Given that** I am a user of the web portal on the home page,
**When** I click on ‘Explore the catalog’ button 
**Then** I should be redirected to ‘/catalog’

### Scenario: Internationalization
**Given that** I am a user of the web portal on the home page,
**When** the selected language is English
**Then** texts should be displayed in English language,
`Explore`
`Data Space`
`Dataset`
`Service`
`See Details`
`Explore the catalog`
**And** metadata of the resources should be displayed in English

**Given that** I am a user of the web portal on the home page,
**When** the selected language is French
**Then** texts should be displayed in French language,
`Explorer`
`Data Space`
`Données`
`Service`
`Voir Détails`
`Explorer le catalogue`
**And** metadata of the resources should be displayed in French 

**Given that** I am a user of the web portal on the home page,
**When** the selected language is German
**Then** texts should be displayed in German language,
`Erkunden Sie`
`Datenraum`
`Datensatz`
`Dienstleistung`
`Siehe Details`
`Den Katalog erforschen`
**And** metadata of the resources should be displayed in German

### Scenario: Mobile display
**Given that** I am a user of the web portal,
**And** I am using a mobile device
**When** I am on the home page
**Then** I should see the mobile version on the catalogue preview (See mock up 2)

## Requirements
Faked data until we fetch them from the ontological database : 

Card 1 : 

* Title : Rhizome
* Category : Agriculture, Food, Environment and Forestry
* Label : Data Space

Card 2 : 

* Title : Agreste 2020
* Category : Agriculture, Food, Environment and Forestry
* Label : Dataset

Card 3 :

* Title : Graphical Plot Registry France 2020
* Category : Agriculture, Food, Environment and Forestry
* Label : Dataset

Card 4 : 

* Title : Data connector
[...]

PR #123, opened 2023-03-22
Feat/add home page explore

AnjaMis:
This PR adds "explore" block on the homepage. At the moment the data inserted is fake and is not translated. The buttons on the cards are also disabled.
- #77 

<details>
<summary>1440px</summary>

</details>

<details>
<summary>768px</summary>

</details>

<details>
<summary>375px</summary>
127 0 0 1_5173_(iPhone 6_7_8)

</details>


src/component/button/button.scss
```diff
@@ -56,6 +56,26 @@
       }
     }
 
+    &.catalog {
+      border: 1px solid themed('secondary-color');
+      color: themed('primary-text');
+      border-radius: 12px;
+      background: themed('secondary-color');
+      box-shadow: 0 4px 12px rgba(255, 149, 0, 0.6);
+
```
lolottetheclash, line 65: you can remove this empty line here

src/component/button/button.scss
```diff
@@ -56,6 +56,26 @@
       }
     }
 
+    &.catalog {
+      border: 1px solid themed('secondary-color');
+      color: themed('primary-text');
+      border-radius: 12px;
+      background: themed('secondary-color');
+      box-shadow: 0 4px 12px rgba(255, 149, 0, 0.6);
```
lolottetheclash, line 64: you can add this color `rgba(255, 149, 0, 0.6)` to the palette

src/component/button/button.scss
```diff
@@ -56,6 +56,26 @@
...
+
+      transition: background-color 200ms ease-in-out;
```
lolottetheclash, line 66: as this transition is used twice in the app, I think you can move it into the `mixins` file to avoid code repetition

src/component/button/button.scss
```diff
@@ -56,6 +56,26 @@
       }
     }
 
+    &.catalog {
```
lolottetheclash, line 59: The `button` component is a generic component that should not contain any "catalog" tinted class or other but only `primary`, `secondary` or other classes being compliant with the design system. 
You can check on the design system that the css code you add is the one corresponding to the `primary` variant button. You should change this to keep the `button` generic ;)

src/component/button/button.scss
```diff
@@ -75,6 +95,11 @@
         gap: 10px;
       }
 
+      &.catalog {
```
lolottetheclash, line 98: same remark here about `catalog`

src/component/button/button.scss
```diff
@@ -89,6 +114,11 @@
           @include noto-sans(400);
           font-size: 16px;
         }
+
+        &.catalog {
```
lolottetheclash, line 118: and same here also

src/component/button/button.tsx
```diff
@@ -11,7 +11,7 @@ type ButtonProps = {
     startIcon?: JSX.Element
     endIcon?: JSX.Element
   }
-  variant?: 'primary' | 'secondary' | 'tertiary'
+  variant?: 'primary' | 'secondary' | 'tertiary' | 'catalog'
```
lolottetheclash, line 14: same remark here

lolottetheclash (changes requested):
Some improvements to make on the `button` component to keep it generic :)

src/component/button/button.scss
ErikssonJoakim: I found out we can just add the primary color in the rgb function so I didn't need to add another color

fredericvilcot (approved):
Great job

PR #135, opened 2023-03-27
Fix/no scroll on small screen

ErikssonJoakim:
A fix to the problem raised  in 
- #117 

This PR makes the sidebar have a scrollbar when content goes out of view. It also fixes the page content not being scrollable when the burger menu is opened


lolottetheclash (approved):
it's better now !

ccamel (approved):
LGTM!

Issue #132, opened 2023-03-24
Bug: Mobile display - tool bar disappears when scroll


MdechampG:
**Describe the bug**

When I use a mobile phone, and I am on any page of the portal, I scroll down to see the components at the bottom of the page. 
These components are not displayed entirely. So I try to scroll down again and then I see the component entirely but the tool bar disappears. 

**To Reproduce**

1. Go to 'app.staging.okp4.network' with a mobile phone
2. Scroll down to the bottom of the page

**Expected behavior**

The components at the bottom of the page should appear entirely and the tool bar should not move. 

**Screenshots**

![IMG_8438.PNG](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/fe3dff9d-e043-41a7-9a3b-5011090416ae)

![IMG_8439.PNG](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/db843450-a035-429f-9273-830146ebd2e9)

**Smartphone (please complete the following information):**

- Device: iPhone 12
- OS: iOS 16.2
- Browser Safari

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206528089) by [Unito](https://www.unito.io)

MdechampG:
Test 29/03

    Device: iPhone 12
    OS: iOS 16.2
    Browser Safari

I still have the bug, even if I refresh or use a private navigation 😞

PR #137, opened 2023-03-28
fix(app): toolbar disappears on mobile scroll

ErikssonJoakim:
Fixes the bug raised in
- #132 

The bug is due to the fact that on mobile the address bar is counted as part of the viewport, making 100vh larger than the actual window. For this reason it is recommended to avoid using 100vh: https://dev.to/admitkard/mobile-issue-with-100vh-height-100-100vh-3-solutions-3nae

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,7 +10,7 @@ const App: FC = () => {
   const isSidebarExpanded = useAppStore(store => store.isSidebarExpanded)
 
   return (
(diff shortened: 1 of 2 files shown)


0xLodz (approved):

lolottetheclash (approved):
👍

ccamel (approved):
LGTM

Issue #88, opened 2023-03-09
Feat : Catalog – Resource Filters  


henigh:
# Feat : Catalog – Resource Filters  

> **As a** user of the portal,
> **I want** to be able to filter by dataverse item (Dataspace // Dataset // Service) 
> **So that** I can have be more specific about what I'm looking for

### Scenario: default display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the dataverse page, 
**Then** I should be able to use a filter that can be applied to the type of dataverse item (Data Space or Dataset or Service)

### Scenario: mobile display 

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the dataverse page, 
**And** I use a mobile device
**Then** I should be able to see and interact with filters in a mobile-friendly way.

### Scenario: Interacting with filters

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I click on `all`
**Then** All types of `dataverse items` should appear
**When** I click on `Data Spaces`
**Then** Only `Data Spaces` will appear
**When** I click on `Datasets`
**Then** Only `Datasets` will appear
**When** I click on `Services`
**Then** Only `Services` will appear
**When** I click on 2 types
**Then** Only cards of these two types will appear
**When** I deselect 1 specific type
**Then** `all` is selected automatically

### Mockups
Desktop : 
![image.png](https://images.zenhubusercontent.com/617955902b89e7481066da3f/4c0e2a7c-f5d9-4bb3-81f7-7e5309f5cbbf)

Mobile : 
![image.png](https://images.zenhubusercontent.com/617955902b89e7481066da3f/82e39254-4dcc-4eb1-b123-a01c338814a2)

### Environment

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206542062) by [Unito](https://www.unito.io)

MdechampG:
Test 24/03/23

Desktop : Firefox, Chrome, Safari
Mobile : Safari 

Scenario : Default display : OK ✅

lolottetheclash:
As seen with @SpicySalm0n @MdechampG & @ErikssonJoakim, the filters are made to wrap on the next line on the small screen instead of staying on one line (as originally planned by the
 original figma), and they can contain text on 2 lines maximum, with an ellipse in case of longer text.
 
 It will now look like that:
 
[...]

PR #139, opened 2023-03-28
fix(chip): text wrapping on small screen

ErikssonJoakim:
Fixing the text wrap issue raised in 
- #88 

<details>
<summary>375px</summary>

Capture d’écran 2023-03-29 à 10 28 50

</details>
<details>
<summary>360px</summary>

Capture d’écran 2023-03-29 à 10 28 34

</details>

OBS I've added extra text in the images to show the behaviour of the chip component when the text overflows.

### DIFF
--- a/src/component/chip/chip.scss
+++ b/src/component/chip/chip.scss
@@ -22,7 +22,6 @@
       display: flex;
       justify-content: space-between;
       align-items: center;
-      white-space: pre;
 
       &:not(.okp4-dataverse-portal-chip-icon) {
         justify-content: center;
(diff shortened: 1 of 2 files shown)


lolottetheclash (changes requested):
You have to dig a little deeper because if you change languages the same problem occurs...
Do we have to allow text wrapping, do we have to put an ellipse on it or should we allow the chip to grow...? I think you can check it with @MdechampG 

Capture d’écran 2023-03-29 à 08 45 58

ErikssonJoakim:
@lolottetheclash Yes I should have accounted for this. I've modified the PR to better handle different text lengths after discussions with @MdechampG and @SpicySalm0n. The filtres now wrap when there is not enough space and for smaller mobile screens they will take the full with. Please see the example images in the PR description

lolottetheclash (approved):
Much better ;)

ccamel (approved):

0xLodz (approved):

Issue #86, opened 2023-03-09
Feat : Data Space Detail Page – tags 


henigh:
# Feat : Data Space Detail Page – tags 

> **As a** user of the portal,
> **I want** to check Sata Space related tags
> **So that** I can quickly understand what this Data Space is about

### Scenario: default display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a Data Space, 
**Then** I should see the related tags appear as presented in the mockup

### Scenario: mobile display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a Data Space, 
**And** I use a mobile device
**Then** I should see the related tags appear in a mobile friendly display

## Technical Specifications
Tags should be fetched from ontology 
Tags can contain text on 2 lines maximum, with an ellipsis in case of longer text (same behavior as Chips)

### Mockups
Desktop : 
![image.png](https://images.zenhubusercontent.com/617955902b89e7481066da3f/f1cae478-5a8e-4328-86f7-3c6b3ebeba9a)

Mobile : 
![image.png](https://images.zenhubusercontent.com/617955902b89e7481066da3f/734e1ab4-f8e0-4448-a698-c1b7ee0eefb3)

### Environment

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206541976) by [Unito](https://www.unito.io)

MdechampG:
Test 06/04/23

Desktop : Firefox, Safari, Chrome
Mobile : Safari (iPhone 12)

Scenario : general display (desktop) : OK ✅ 
Note : I can't test when the tag is long with these faked data -> To test with the ontology connected

Scenario : mobile display : OK ✅ 

Thanks @0xLodz !

MdechampG:
You're right :) It's good thank you ! 

![image](https://user-images.githubusercontent.com/92780073/230345908-0a825890-4a92-48bb-b60d-d0ceae33e0cd.png)

Issue #71, opened 2023-03-01
Feat : General description for a Dataspace


henigh:
# Feature: General description for a Data Space

> **As a** user of the portal,
> **I want** to see a Data Space description 
> **So that** I can check out the purpose of the Data Space before joining it.

### Scenario: default display (desktop)

**Given that** I am a user of the web portal (unauthenticated or authenticated),
**And** I am using a *desktop* device,
**When** I am on the Data Space detail page 
**Then** I should see the title of the Data Space, its label (Data Space) and a proper description that appear as presented in the mockup below

### Scenario : Mobile Display
**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on a Data Space detail page
**And** I use a mobile device
**Then** I should see mobile friendly display (See mock up)

### Scenario : Internationalization
**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** the selected language is English, French, or German
**Then** the `metadata` should be displayed in the adequate selected language (See config)

## Resources

Ontology examples : https://github.com/okp4/ontology/tree/main/example/rhizome

### Mockups

Desktop. 

![image.png](https://images.zenhubusercontent.com/617955902b89e7481066da3f/8b74ec30-6f91-43e8-8000-23ebbfc30484)

### Mobile
<img width="251" alt="image" src="https://user-images.githubusercontent.com/39368385/223369700-00c09aa1-fe5f-44cc-8962-ff5dd92e2c8b.png">

### Environment

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206531549) by [Unito](https://www.unito.io)

MdechampG:
Test 07/04
Desktop : Firefox and Safari 
Mobile : Safari

Scenario : Default Display : OK ✅ 

Scenario : Mobile Display : OK ✅ 

Scenario : Internationalization: OK ✅ 
Nb : Impossible to test the internationalization of the metadata as the ontology is not yet connected.

Issue #68, opened 2023-02-28
Feat : Resource Detail - general description


MdechampG:
# Feat : Resource Detail - general description

> **As a** user of the portal,
> **I want** to read a general description of a shared resource (data or service)
> **So that** I can understand what this resource is about, its context of creation and its purpose. 

### Scenario: default display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a resource, 
**Then** I should see the name of the resource, its label (type of the resource) and a general description (see mock up)

### Scenario: mobile display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a resource, 
**And** I use a mobile device
**Then** I should see the name of the resource, its label (type of the resource) and a general description (see mock up mobile device)

### Scenario: Internationalization

**Given that**I am a user of the web portal (authenticated or unauthenticated),
**When** the selected language is English
**Then** texts should be displayed in English language,
`Data Space`
`Dataset`
`Service`
**And** metadata of the resources should be displayed in English

**Given that** I am a user of the web portal on the home page,
**When** the selected language is French
**Then** texts should be displayed in French language,
`Data Space`
`Données`
`Service`
**And** metadata of the resources should be displayed in French 

**Given that** I am a user of the web portal on the home page,
**When** the selected language is German
**Then** texts should be displayed in German language,
`Datenraum`
`Datensatz`
`Dienstleistung`
**And** metadata of the resources should be displayed in German

### Mockups
Desktop : 
![image.png](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/6fd28c4f-d168-47a3-ad82-27c0d8426bbd)

Mobile : 
![image.png](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/d465cfcf-3610-4198-8394-63cb8783a864)

### Environment

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206531434) by [Unito](https://www.unito.io)

MdechampG:
Test 07/04
Desktop : Firefox and Safari 
Mobile : Safari

Scenario : Default Display : OK ✅ 

Scenario : Mobile Display : OK ✅ 

Scenario : Internationalization: OK ✅

PR #140, opened 2023-03-28
Feat/dataverse description && tags

GeoffreyPicard:
This PR implements:
https://github.com/okp4/dataverse-portal/issues/86
https://github.com/okp4/dataverse-portal/issues/71
https://github.com/okp4/dataverse-portal/issues/68

Capture d’écran 2023-03-30 à 09 48 13

### DIFF
--- a/src/component/tag/tag.scss
+++ b/src/component/tag/tag.scss
@@ -1,26 +1,27 @@
 @import '@/style/fonts.scss';
 @import '@/style/theme.scss';
 
-$colors: (
-  'primary-color': 'background-primary-variant',
-  'primary-color-variant-3': 'background-primary-variant-3',
-  'primary-color-variant-4': 'background-primary-variant-4'
-);
-
 .okp4-dataverse-portal-tag-main {
   @include with-theme {
     @include noto-sans(700);
-    border-radius: 0 20px;
-    padding: 8px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    color: themed(primary-color-variant-3);
     font-size: 14px;
-    line-height: 24px;
-    letter-spacing: 2px;
+    min-height: 34px;
+    padding: 0 8px;
+    line-height: 18px;
+    background: rgba(themed(primary-color-variant-3), 0.12);
+    border-radius: 6px;
 
-    @each $color-key, $color-value in $colors {
-      &.#{$color-key} {
-        color: themed($color-key);
-        background-color: themed($color-value);
-      }
+    .okp4-dataverse-portal-tag-name {
+      display: box;
+      -webkit-line-clamp: 2;
+      -webkit-box-orient: vertical;
+      text-overflow: ellipsis;
+      overflow: hidden;
+      max-width: 240px;
     }
   }
 }
--- a/src/component/tag/tag.tsx
+++ b/src/component/tag/tag.tsx
@@ -1,13 +1,12 @@
-import classNames from 'classnames'
+import type { FC } from 'react'
 import './tag.scss'
 
-export type ColorVariant = 'primary-color' | 'primary-color-variant-3' | 'primary-color-variant-4'
-
 type TagProps = {
-  label: string
-  colorVariant?: ColorVariant
+  tagName: string
 }
 
-export const Tag = ({ label, colorVariant = 'primary-color-variant-3' }: TagProps): JSX.Element => (
-  <div className={classNames('okp4-dataverse-portal-tag-main', colorVariant)}>{label}</div>
+export const Tag: FC<TagProps> = ({ tagName }) => (
+  <div className="okp4-dataverse-portal-tag-main">
+    <div className='okp4-dataverse-portal-tag-name'>{tagName}</div>
+  </div>
 )
--- a/src/component/card/dataverseItemCard/dataverseItemCard.scss
+++ b/src/component/card/dataverseItemCard/dataverseItemCard.scss
@@ -2,7 +2,13 @@
 @import '@/style/mixins.scss';
 @import '@/style/theme.scss';
 
-.okp4-dataverse-portal-dataverse-item-card-main {
+$colors: (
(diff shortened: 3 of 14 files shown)


src/component/tag/tag.scss
```diff
@@ -1,26 +1,17 @@
...
-}
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: $primary-variant-3;
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 22px;
+  font-family: 'Noto Sans';
+  height: 34px;
```
ErikssonJoakim, line 13: IMO the tag dimension should be set by its parent

src/component/tag/tag.scss
```diff
@@ -1,26 +1,17 @@
...
+  padding: 0 8px;
```
ErikssonJoakim, line 14: Missing padding top and bottom

ErikssonJoakim (changes requested):
Great, I just have some minor changes and suggestions that I would like you to take into consideration

src/component/tag/tag.scss
0xLodz: We would've been more explicit by writing `padding: 0 8px 0 8px`. The `padding` shorthand property can be simplified when both top/bottom or left/right are similar. In that case `padding: 0 8px` is equivalent to `padding: 0 8px 0 8px`

src/component/tag/tag.scss
```diff
@@ -1,26 +1,17 @@
...
-        color: themed($color-key);
-        background-color: themed($color-value);
-      }
-    }
-  }
-}
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: $primary-variant-3;
```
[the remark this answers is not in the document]
GeoffreyPicard: Yes I do it

src/component/tag/tag.scss
GeoffreyPicard: Normally component from design system must be independent. 
We can also add variables like "size" or  "fullWidth" callable by parent

src/component/tag/tag.scss
GeoffreyPicard: I used fix height so we dont need to add padding top and bot.
If the problem is the case of two lines, we can block a number of characters and use "..." at the end in CSS.
But normally we don't want a tags in two lines :)

ErikssonJoakim: Fair point 👍

ErikssonJoakim (approved):
LGTM 😃

fredericvilcot:
@GeoffreyPicard 
Before reviewing this PR, I'd like to better understand what do you mean in the main description (and in your commit https://github.com/okp4/dataverse-portal/pull/140/commits/cc09157f05003be198ed410d2229bd0186929ca3) when you say "_and create a new Tag component (from design system)._".
What do you mean by design system?
Thx

src/component/tag/tag.scss
```diff
@@ -1,26 +1,19 @@
...
-    letter-spacing: 2px;
-
-    @each $color-key, $color-value in $colors {
-      &.#{$color-key} {
-        color: themed($color-key);
-        background-color: themed($color-value);
-      }
-    }
+    line-height: 22px;
+    font-family: 'Noto Sans';
```
lolottetheclash, line 13: You should use the font mixin here ;)

src/component/tag/tag.tsx
```diff
@@ -1,13 +1,12 @@
...
 type TagProps = {
-  label: string
-  colorVariant?: ColorVariant
+  tagName: string
 }
 
-export const Tag = ({ label, colorVariant = 'primary-color-variant-3' }: TagProps): JSX.Element => (
-  <div className={classNames('okp4-dataverse-portal-tag-main', colorVariant)}>{label}</div>
-)
+export const Tag: FC<TagProps> = ({ tagName }) => {
```
lolottetheclash, line 8: You can remove useless brackets and `return` here

src/component/card/dataverseCard/dataverseCard.tsx
```diff
@@ -27,6 +28,8 @@ export const DataverseCard: FC<DataverseCardProps> = ({ type, label, description
         return 'primary-color-variant-3'
       case 'dataset':
         return 'primary-color-variant-4'
+      default:
```
lolottetheclash, line 31: As the type of `DataverseItem` is one of `'dataspace' | 'dataset' | 'service'` we don't need a default value here because every case is handled.

lolottetheclash (changes requested):
Thanks for this PR! I let you some remarks, I also think that the behavior of the tag should be planned if the text to be displayed is too long for its container. We 'll discuss about it!

GeoffreyPicard:
> @GeoffreyPicard Before reviewing this PR, I'd like to better understand what do you mean in the main description (and in your commit [cc09157](https://github.com/okp4/dataverse-portal/commit/cc09157f05003be198ed410d2229bd0186929ca3)) when you say "_and create a new Tag component (from design system)._". What do you mean by design system? Thx


By "design system" I mean components from there:
https://www.figma.com/file/QWGxMd0sVIpvNVv5GZgsmI/OKP4---Design-System?node-id=1-2&t=oia41A6W44VimLiL-0

src/component/card/dataverseCard/dataverseCard.tsx
```diff
@@ -39,7 +42,7 @@ ${description}`,
   return (
     <Card>
       <div className="okp4-dataverse-portal-dataverse-card-main">
-        <Tag colorVariant={renderTagColor(type)} label={t(`resources.${type}`)} />
+        <div className={classNames('okp4-dataverse-portal-dataverse-card-category', renderTagColor(type))}>{t(`resources.${type}`)}</div>
```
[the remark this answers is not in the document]
GeoffreyPicard: Good idea I change it

Issue #87, opened 2023-03-09
Feat : Data Space Detail Page – topics


henigh:
# Feat : Data Space Detail Page – topics 

> **As a** user of the portal,
> **I want** to check the topics related to one Data Space
> **So that** I can quickly understand what this Data Space is about (topics covered by the Data Space)

### Scenario: default display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a Data Space, 
**Then** I should see the related topics appear as presented in the mockup

### Scenario: mobile display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a Data Space, 
**And** I use a mobile device
**Then** I should see the related topics appears in a mobile friendly display

### Scenario: Internationalization
Topics should appear **correctly** regarding the selected language.

**EN**

> - Agriculture, Food, environment and forestry.
> - Biology, geology, and chemistry
> - Industry, mobility, and engineering
> - Logistics, retail, supply chain, and eCommerce
> - DeFi and Crypto
> - Healthcare
> - Business and purchase
> - Marketing and Customer Behavior
> - Energy
> - Other

**FR**
> - Agriculture, alimentation, environnement et foresterie.
> - Biologie, géologie et chimie
> - Industrie, mobilité et ingénierie
> - Logistique, commerce de détail, chaîne d'approvisionnement et eCommerce
> - Crypto et finance décentralisée
> - Soins de santé
> - Commerce et approvisionnement 
> - Marketing et comportement des clients
> - Énergie
> - Autres

**GR**
> - Landwirtschaft, Ernährung, Umwelt und Forstwirtschaft
> - Biologie, Geologie und Chemie
> - Industrie, Mobilität und Technik
> - Logistik, Einzelhandel, Lieferkette und eCommerce
> - DeFi und Krypto
> - Gesundheitswesen
> - Wirtschaft und Einkauf
> - Marketing und Kundenverhalten
> - Energie
> - Sonstiges

## Technical Specifications
topics should be fetched from ontology 

### Mockups
Desktop : 
![image.png](https://images.zenhubusercontent.com/617955902b89e7481066da3f/bbc0cbf2-33f1-4cac-8b3d-ef9aace289d0)

Mobile : 
![image.png](https://images.zenhubusercontent.com/617955902b89e7481066da3f/15c88b4e-587a-4cd1-bf55-b1fc420d8d41)

### Environment

[...]

Issue #69, opened 2023-03-01
Feat : Resource Detail - General metadata


MdechampG:
# Feat : Resource Detail - General metadata

> **As a** user of the portal,
> **I want** to read general metadata of a shared resource (data or service)
> **So that** I can understand better what is the resource and its properties.  

### Scenario: default display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a resource, 
**Then** I should see the general metadata of the resources (See mock up)

### Scenario: mobile display (desktop)

**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** I am on the detail page of a resource, 
**And** I use a mobile device
**Then** I should see the general metadata of the resources on its mobile version (See mock up)

### Scenario : Temporal Coverage

**Given that** I am a user of the web portal (authenticated or unauthenticated) on the detail page of a resource
**When** the metadata `Start date` and `End date`are fulfilled in the ontology,
**Then** I should see `Year of Start Date` - `Year of End Date` (Mock up 1) 
**When** the metadata `Start date` is fulfilled in the ontology and `End Date` is missing
**Then** I should see `From Year of Start Date` (Mock up 2) 
**When** the metadata `End date` is fulfilled in the ontology and `Start Date` is missing
**Then** I should see `To Year of End Date` (Mock up 3) 
**When** Start date and End date are missing
**Then** Temporal Coverage is not displayed at all

### Scenario: Internationalization

**Given that**I am a user of the web portal (authenticated or unauthenticated),
**When** the selected language is English
**Then** texts should be displayed in English language,
`Format`
`Topic`
`Licence`
`Geographical coverage`
`Temporal coverage`
**And** metadata of the resources should be displayed in English

**Given that** I am a user of the web portal on the home page,
**When** the selected language is French
**Then** texts should be displayed in French language,
`Format`
`Domaine`
`Licence`
`Couverture géographique`
`Couverture temporelle`
**And** metadata of the resources should be displayed in French 

**Given that** I am a user of the web portal on the home page,
**When** the selected language is German
**Then** texts should be displayed in German language,
`Format`
`Thema`
`Lizenz`
`Geografischer Geltungsbereich`
`Zeitlicher Geltungsbereich`
**And** metadata of the resources should be displayed in German

#### Desktop : 

Dataset : 
1/
[...]

PR #138, opened 2023-03-28
Feat: add dataverse items dynamic routes

lolottetheclash:
The PR implements this [issue](https://github.com/okp4/dataverse-portal/issues/97).
It enables clic on dataverse card buttons located in Home and in Explore pages.
For the moment, if the URL id param doesn't exist, a message "Dataspace/Dataset/Service is not found" appears while waiting for the layout of the 404 page. 
🤔 The data as well as the way of querying these data is going to change, a refactor will be done at this time to decouple the elements.

PR #141, opened 2023-03-28
Feat: Resource detail component - General metadata

0xLodz:
This PR implements https://github.com/okp4/dataverse-portal/issues/69 & https://github.com/okp4/dataverse-portal/issues/87.

Local testings can be performed on dataverse item pages: `/dataverse/[dataverseItem]/[id]`
Example for the screenshot below: `http://127.0.0.1:5173/dataverse/dataset/1`

image

## Fallback to `-` when Temporal Coverage metadata value is not an ISOstring.
image


fredericvilcot:
@0xLodz 
Please, rebase your branch from `origin/main` before opening a PR, because this PR embeds commits from another branches and does not target `main`.
Thx

fredericvilcot:
@0xLodz 
I am a little bit confused about this PR, you say it implements #69, but how can we test the implementation requested in the feature as your component is imported nowhere?
If this PR aims to implement the feature, it should match the acceptance criteria asked from the product teal, that is obviously not the case.
Thanks .

src/component/resource-details/resourceDetails.tsx
```diff
@@ -0,0 +1,20 @@
+import './resourceDetails.scss'
+import type { FC } from 'react'
+
+type ResourceDetailsProps = {
+  icon: React.ReactNode
```
ccamel, line 5: https://dev.to/ridhoanshory/jsx-element-react-node-react-element-idi

0xLodz:
> @0xLodz I am a little bit confused about this PR, you say it implements #69, but how can we test the implementation requested in the feature as your component is imported nowhere? If this PR aims to implement the feature, it should match the acceptance criteria asked from the product team, that is obviously not the case. Thanks .

@fredericvilcot sorry for the confusion. This feature has the PR #138 as dependency, that's why I initially set [feat/add-dataverse-items-dynamic-routes](https://github.com/okp4/dataverse-portal/tree/feat/add-dataverse-items-dynamic-routes) that implements #138 as the target branch, in order to provide only commits related to this feature. After reading you [previous message](https://github.com/okp4/dataverse-portal/pull/141#issuecomment-1487479202), I changed the target branch and removed the commit that provided a way to test the feature. In this case I don't see any other solution than changing the target branch from `main` to [feat/add-dataverse-items-dynamic-routes](https://github.com/okp4/dataverse-portal/tree/feat/add-dataverse-items-dynamic-routes). What would you recommend?
Thanks.

fredericvilcot:
@0xLodz 
To keep a clean and consistant history, we need to wait for the merge of the blocking pull requests.
Please switch this one to draft meanwhile.
Thank you.

src/page/dataverse/component/generalMetadata/generalMetadata.tsx
```diff
@@ -0,0 +1,57 @@
...
+  | 'format'
+  | 'topic'
+  | 'license'
+  | 'geographicalCoverage'
+  | 'temporalCoverage'
+  | 'category'
+
+export type GeneralMetadata = {
+  iconName: IconName
+  title: GeneralMetadataTitle
```
ErikssonJoakim, line 19: We cannot specify the different titles in the component because the title value depends on the caller. For example it makes no sense for the data space page to have the format title because a data space does not have a format meta data

ErikssonJoakim (changes requested):
Great work 😃 There are some things to consider but it's a good start!

src/page/dataverse/dataset/dataset.tsx
```diff
@@ -1,9 +1,50 @@
...
+    iconName: 'calendar',
+    title: 'temporalCoverage',
+    description: '2020'
+  }
+]
+
+type MatchedDatasetProps = {
+  label: string
+}
+const MatchedDataset: FC<MatchedDatasetProps> = ({ label }) => (
```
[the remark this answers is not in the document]
0xLodz: I do agree that this abstraction is unnecessary, it add some noise in the file. We discussed about it with @lolottetheclash. I did that by providing more readability at the `match` function level. I am ok to remove it. What do you think @lolottetheclash ?

src/page/dataverse/component/generalMetadata/generalMetadata.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export type GeneralMetadataTitle =
+  | 'format'
+  | 'topic'
+  | 'license'
+  | 'geographicalCoverage'
+  | 'temporalCoverage'
+  | 'category'
+
+export type GeneralMetadata = {
```
ccamel, line 17: There are three distinct GeneralMetadata models: one for Dataspace, one for Service, and one for Dataspace. While it may seem possible to merge them into one model, doing so would result in inaccuracies in the meaning and nature of the conveyed information. Therefore, it is advisable to rely on the concepts of the ontology for consistency as much as possible.

src/page/dataverse/dataset/dataset.tsx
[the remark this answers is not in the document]
lolottetheclash: @0xLodz I think that this decision is personal and that we do not all share the same point of view, I will say that it is not decisive, up to you ;)

src/page/dataverse/component/generalMetadata/generalMetadata.tsx
ccamel: Definitively the second option. But, please take a step back, and consider the element to be displayed as a metadata property (and not a general metadata title), which describes a property (what you call the title), its value (what you call description), to which you associate an icon, eventually:

```ts
type MetadataProperty = {
  property: string
  value: string
  iconName?: IconName
}
```

You may not know this yet, but ontology properties, like the format property (`core:hasFormat`) are also described as such, and therefore have a label property (`rdfs:label`), internationalized, which specifies the label to use when representing the property. This is why you can't use: `t(`generalMetadata.${title}`)`. So it makes sense that the property is a string, since the value conveys the exact representation of the property, for the current local language.

Now, another issue arises with the values. You are assuming that it is a string, but that is not necessarily the case. For instance, the temporal coverage is a date range, and the dates must be represented according to the locale. And this is the responsibility of the component, imho... This is just fantastic 😅

src/ui/view/dataverse/component/generalMetadata/generalMetadata.tsx
```diff
@@ -0,0 +1,54 @@
...
+import './generalMetadata.scss'
+import './i18n/index'
+
+export type MetadataProperty = {
+  property: string
+  value: string
+  iconName?: IconName
+}
+
+const GeneralMetadataItem: FC<MetadataProperty> = memo(
```
ccamel, line 17: Can you explain the purpose of `memo` here. Thanks :pray:

src/util/isoDateTime/isoDateTime.ts
```diff
@@ -0,0 +1,15 @@
+const getYearFromISODateTime = (ISODateTime: string, locale: string): string => {
+  const date = new Date(ISODateTime)
+  const formatter = new Intl.DateTimeFormat(locale, { year: 'numeric' })
+  return formatter.format(date)
+}
```
ccamel, line 5: In the function, I would delegate the responsibility of determining whether a string is an ISO DateTime to the `Date` constructor. Moreover, as it stands, if you pass a string that does not comply with the format, the function throws an exception.

ccamel (changes requested):

src/ui/page/dataverse/dataset/dataset.tsx
```diff
@@ -15,7 +45,14 @@ const Dataset = (): JSX.Element => {
 
   return match(
     () => <p>Dataset not found</p>,
-    (dataset: DataverseItemDetails) => <p>{dataset.label}</p>
+    (dataset: DataverseItemDetails) => {
+      return (
```
fredericvilcot, line 49: Please keep ES6 syntax

src/ui/page/dataverse/dataspace/dataspace.tsx
```diff
@@ -15,7 +25,14 @@ const Dataspace = (): JSX.Element => {
 
   return match(
     () => <p>Dataspace not found</p>,
-    (dataspace: DataverseItemDetails) => <p>{dataspace.label}</p>
+    (dataspace: DataverseItemDetails) => {
+      return (
```
fredericvilcot, line 29: Please keep ES6 syntax

src/ui/page/dataverse/service/service.tsx
```diff
@@ -15,7 +25,14 @@ const Service = (): JSX.Element => {
 
   return match(
     () => <p>Service not found</p>,
-    (service: DataverseItemDetails) => <p>{service.label}</p>
+    (service: DataverseItemDetails) => {
+      return (
```
fredericvilcot, line 29: Please keep ES6 syntax

src/ui/view/dataverse/component/generalMetadata/generalMetadata.tsx
0xLodz: In that case it avoids unnecessary re-renders of the `GeneralMetadataItem` by memoizing it's output according to its props, If it's parent re-render for any reason, it won't be re-rendered. However this component is quite small, it might be worse to apply this optimisation mostly on components that are expensive to render.

src/util/isoDateTime/isoDateTime.ts
fredericvilcot: Yes, let's delegate the responsibility to the Date constructor and treat returned value (or Error) before calling `convertToYearIfISODateTime` method: it will prevent from returning a non formatted input.

src/util/isoDateTime/isoDateTime.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+    ['2023-01-01', false],
+    ['2023-01-01T00:00', false],
+    ['2023-01-01T00:00:00', true],
+    ['2023-01-01T00:00:00.000', true],
+    ['2023-01-01T00:00:00.000Z', true],
+    ['2023-01-01T00:00:00.000+01:00', true],
+    ['2023-01-01T00:00:00.000-01:00', true],
+    ['002023-01-01T00:00:00.000+01:00', false]
+  ])('validates input %s as %s', (input: string, expected: boolean) => {
+    expect(isISODateTime(input)).toBe(expected)
```
fredericvilcot, line 16: IMHO you should only test the `convertToYearIfISODateTime` method that will test all other methods in a traversal way and ensures you 100% of coverage.

fredericvilcot (changes requested):
Great job, I have some remarks to be discussed, thx

src/util/isoDateTime/isoDateTime.ts
0xLodz: What do you think about a validation that use both the regex and the `Date` constructor to determine if a string is an ISODateTime? Because for example `new Date(1)` is a date.

> Moreover, as it stands, if you pass a string that does not comply with the format, the function throws an exception.
You are right, that the reason why I didn't exported the function. That's probably too implicit and needs error handling if we want to use it directly.

src/util/isoDateTime/isoDateTime.spec.ts
0xLodz: Thanks for pointing this out. That makes sense to get a full coverage.

src/util/isoDateTime/isoDateTime.ts
```diff
@@ -0,0 +1,15 @@
...
+export const isISODateTime = (input: string): boolean => {
+  const isoDateTimeRegex: RegExp =
+    /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(\.\d{1,6})?(([+-])(\d{2}):(\d{2})|Z)?$/
+
+  return isoDateTimeRegex.test(input) && new Date(input).toString() !== 'Invalid Date'
+}
+
+export const convertToLocalizedYearIfISODateTime = (input: string, locale: string): string => {
+  return isISODateTime(input) ? getLocalizedYear(new Date(input), locale) : input
+}
```
fredericvilcot, line 15: You should not return the `input`: what about if I give `foo` as the `input` parameter of the `convertToLocalizedYearIfISODateTime` method? => the view will display `foo` as a temporal coverage, that is a wrong interpretation.
In a functional way, we would have used a `Either` task to capture both values (result, error) and then handle it in the view.
For now, I'll suggest either to throw an error or return a `Invalid date` that will be directly handle as a fallback in the view. (content must be validated with the product team)

fredericvilcot (changes requested):

src/util/isoDateTime/isoDateTime.ts
fredericvilcot: And please, keep ES6 syntax 😉

ccamel: @0xLodz The point is: _never trust the input_. The function `convertToLocalizedYearIfISODateTime` should function consistently in _any_ scenario, with the following expected behavior:

- It should return the year of the input date if it can be successfully parsed
- If it cannot be parsed, an error should be returned. The approach to return the error is at your discretion: _functional style_ (the prefered one) using an [Either<>](https://gcanti.github.io/fp-ts/modules/Either.ts.html) value, or _imperative style_ by throwing an error, as suggested by @fredericvilcot

src/util/isoDateTime/isoDateTime.ts
ccamel: Yes. My previous comment was inaccurate. You should rely on the `Date.parse()` function to do what you want without needing a regex. Just check that the number returned is not `NaN`, and it will work as expected.

fredericvilcot (approved):

lolottetheclash (approved):
👌

ccamel (approved):
Let's go! Adjustments will be made when the data model is available

PR #145, opened 2023-03-30
Feat/add import global scss

GeoffreyPicard:
We now use Vite preprocessor to import main scss files globally 
I removed useless import related to this update


fredericvilcot:
@GeoffreyPicard 
I close this PR as it's not relevant for now.

Issue #72, opened 2023-03-01
Feat : Governance description for a Data Space


henigh:
# Feature: Governance description for a Data Space
> **As a** user of the portal,
> **I want** to see a Data Space Governance description 
> **So that** I can quickly check out governance mainlines 

### Scenario 1 : default display (desktop)
**Given that** I am a user of the web portal (unauthenticated or authenticated),
**And** I am using a _desktop_ device,
**When** I am on the Data Space detail page 
**Then** I should see a proper description for the resource governance (Data Space in this particular case) that appears as presented in the mockup below

### Scenario 2 : Access governance button
**Given that** I am a user of the web portal (unauthenticated or authenticated),
**And** I am using a _desktop_ device,
**Then** the button 'See Governance' is disabled

### Scenario 3 : Mobile Display
Given that I am a user of the web portal (authenticated or unauthenticated),
When I am on Data Space detail page 
And I use a mobile device
Then I should see mobile friendly display (See mock up)

### Scenario 4 : Internationalization
**Given that** I am a user of the web portal (authenticated or unauthenticated),
**When** the selected language is **English**, **French**, or **German**
**Then**`title`,  `description` and `button` should be displayed in the adequate selected language _(See config)_

## Config example
As the dedicated service to fetch the Data Space governance description from the blockchain is not ready yet, here is the description for 2 Data Spaces:

### DS4I Data Space
**En**
> Governance _(title)_
> DS4I is a private Data Space where resources are only accessible for a group of addresses contained in a dedicated whitelist.
> Only OKP4 have the authority to edit the Whitelist. _(description)_
> Access governance _(button)_

**Fr**
> Gouvernance _(title)_
> DS4I est un Data Space privé où les ressources ne sont accessibles que pour un groupe
> d'adresses de wallets contenues dans une Whitelist dédiée.
> Seul OKP4 a le droit de modifier cette whitelist. _(description)_
> Accéder à la governance _(button)_

**DE**
> Führung  _(title)_
> DS4I ist ein privater Data Space, in dem die Ressourcen nur für eine Gruppe zugänglich sind.
> Adressen von Wallets, die in einer dedizierten Whitelist enthalten sind.
> Nur OKP4 hat das Recht, diese Whitelist zu ändern. _(description)_
> Auf Governance zugreifen _(button)_

### Rhizome
**En**

> This first Data Space has a centralized governance: only OKP4 can modify the rules.
> In this first version, only OKP4 can register data and services. 
[...]

PR #148, opened 2023-03-30
Feat/button size property

ErikssonJoakim:
This PR adds the `size: 'small' | 'large'` property to the button component, removes wrongly set svg fill as well as adjusts button borders and `label` line-height to correctly reflect the mock-up. This is done in preparation of the following issue
- #72 

Where the secondary button variant require this property (e.g. `DataverseItemCard` uses the small secondary button & the access governance button is a large secondary button).

One thing to take into consideration is that this changes the `gap` in the wallet button from `10px` to `12px`, something that is not according to the mock-up. I've done this for consistency in the property behaviour as I don't see it as a major change.

### DIFF
--- a/src/ui/component/button/button.scss
+++ b/src/ui/component/button/button.scss
@@ -36,13 +36,6 @@
   }
 }
 
-@mixin disabled {
-  cursor: unset;
-  background: themed('neutral-variant-1');
-  border-color: themed('neutral-variant-1');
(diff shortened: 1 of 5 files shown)


ccamel (approved):
LGTM!

lolottetheclash (approved):
Good for me too 👌

Issue #97, opened 2023-03-13
Feat: See details of a resource


MdechampG:
# Feature: See details of a resource
> **As a** user of the portal on the catalog 
> **I want** to click on `See details`
> **So that** I can have more information about a resource registered in the Dataverse

### Scenario: default display (desktop)
**Given that** I am a user of the web portal on the home page,
**When** I click on ‘See details’ button of a resource
**Then** I should be redirected to the resource details page of this resource (See requirements)
**And** I should see an arrow to go on Dataverse page

## Requirements
The URL of the resource detail page should be with the framework : 
Data space : `/dataverse/dataspace/ID`
Dataset : `/dataverse/dataset/ID`
Service : `/dataverse/service/ID`

### Mockups
Mock up  : 
![image.png](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/242dbd7d-998d-4b10-8605-13c2044e190f)

![image](https://user-images.githubusercontent.com/92780073/228909276-f4d2f9e9-8a49-4d47-9ac6-6b74cec973f4.png)

### Environment

* staging URL: <https://app.staging.okp4.network/>

## Additional informations
None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206542387) by [Unito](https://www.unito.io)

ccamel:
@MdechampG it would be advisable to specify the behavior if the text to be displayed is very long (both for the title and the description) - we display everything, we cut, we ellipse...

MdechampG:
@lolottetheclash I have added the spec about the previous page arrow. I put this issue in `To Do` as discussed :)

0xLodz:
@lolottetheclash @MdechampG should we display `Dataverse` according to the mockup ?
![image](https://user-images.githubusercontent.com/8344874/230043068-9313fedc-88a9-4793-84d2-6ee1980d2123.png)
Or will it be done later?

MdechampG:
This new feature is in this issue : https://github.com/okp4/dataverse-portal/issues/168

MdechampG:
Test 07/04
Desktop : Safari, Firefox
Mobile : Safari

Scenario: default display (desktop) OK ✅

PR #161, opened 2023-03-31
Feat: add back to dataverse button on dataverse items pages

lolottetheclash:
This PR adds a previous button which is now displayed on `Dataset`, `Dataspace` and also `Service` page.
This button allows for now to navigate to `Dataverse` page (even if we come from `Home` page).
It completes this [issue](https://github.com/okp4/dataverse-portal/issues/97) and  [this one](https://github.com/okp4/dataverse-portal/issues/168).

image

image

### DIFF
--- a/src/ui/i18n/translations/common_de.json
+++ b/src/ui/i18n/translations/common_de.json
@@ -15,5 +15,6 @@
     "details": "Siehe Details",
     "accessGovernance": "Auf Governance zugreifen"
   },
-  "filters": "Filter"
+  "filters": "Filter",
+  "dataverse": "Dataverse"
 }
--- a/src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
+++ b/src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
@@ -1,4 +1,5 @@
 @import '@/ui/style/mixins.scss';
+@import '@/ui/style/theme.scss';
 
 .okp4-dataverse-portal-dataverse-component-page-template-main {
   @include twelve-column-layout;
@@ -8,5 +9,19 @@
     display: flex;
     flex-direction: column;
     row-gap: 40px;
+
+    .okp4-dataverse-portal-dataverse-button {
+      @include with-theme {
+        width: fit-content;
+        padding: 10px 7px;
+        cursor: pointer;
+
+        svg {
+          path {
+            fill: themed('text');
+          }
+        }
+      }
+    }
   }
 }
--- a/src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
+++ b/src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
@@ -1,5 +1,6 @@
-@import '@/ui/style/mixins.scss';
 @import '@/ui/style/theme.scss';
+@import '@/ui/style/fonts.scss';
+@import '@/ui/style/mixins.scss';
 
 .okp4-dataverse-portal-dataverse-component-page-template-main {
   @include twelve-column-layout;
@@ -10,16 +11,33 @@
     flex-direction: column;
     row-gap: 40px;
 
-    .okp4-dataverse-portal-dataverse-button {
+    .okp4-dataverse-portal-dataverse-back {
       @include with-theme {
-        width: fit-content;
-        padding: 10px 7px;
-        cursor: pointer;
+        display: flex;
+        align-items: center;
 
-        svg {
-          path {
-            fill: themed('text');
+        .okp4-dataverse-portal-dataverse-back-button {
+          display: flex;
+          justify-content: center;
+          align-items: center;
+          background: none;
+          border: 0;
+          width: fit-content;
+          padding: 10px 7px;
+          cursor: pointer;
+  
+          svg {
+            path {
+              fill: themed('text');
+            }
           }
+        }
+        
+        .okp4-dataverse-portal-dataverse-back-text {
+          @include noto-sans(700);
+          font-size: 14px;
+          line-height: 22px;
+          color: themed('text');
         }
       }
     }
--- a/src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
+++ b/src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
@@ -4,42 +4,43 @@
 
 .okp4-dataverse-portal-dataverse-component-page-template-main {
   @include twelve-column-layout;
+  row-gap: 19px;
+
+  .okp4-dataverse-portal-dataverse-back {
+    @include with-theme {
+      display: flex;
+      align-items: center;
+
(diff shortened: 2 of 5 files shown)


0xLodz (commented):
@lolottetheclash  I validate your commit 1430352ed64aa2ecd14131c4920ca25693a262a9 👍 
However I can't approve the PR because it contains commits that have been rewritten (hashes changes) from another PR or are not up to date. Could you please base your work on the branch `feat/dataverse-tags`, thus we quickly see the change you made, without changing base branch commits hashes? Thanks

0xLodz (approved):
Base branch is now [feat/dataverse-tags](https://github.com/okp4/dataverse-portal/tree/feat/dataverse-tags). I am wondering if we merge this one in it, or if we wait for `feat/dataverse-tags` to be marged?

src/ui/view/dataverse/component/pageTemplate/pageTemplate.tsx
```diff
@@ -30,6 +35,12 @@ const PageTemplate: FC<PageTemplateProps> = ({ data, metadata }): JSX.Element =>
   return (
     <div className="okp4-dataverse-portal-dataverse-component-page-template-main">
       <div className="okp4-dataverse-portal-dataverse-page-template-left-side-wrapper">
+        <div className="okp4-dataverse-portal-dataverse-back">
+          <button className="okp4-dataverse-portal-dataverse-back-button" onClick={backToDataverse}>
+            <Icon name="arrow-left" />
+          </button>
+          <span className="okp4-dataverse-portal-dataverse-back-text">Dataverse</span>
```
lolottetheclash, line 42: To keep consistency, could you add "dataverse" to common i18n? even if translation is the same for our 3 proposed languages

src/ui/i18n/translations/common_de.json
```diff
@@ -13,5 +13,6 @@
     "explore": "Erkunden Sie",
     "details": "Siehe Details"
   },
-  "filters": "Filter"
+  "filters": "Filter",
+  "dataverse": "Dataverse"
```
ErikssonJoakim, line 17: Why not just add the text directly in the element since the translations are all the same?

src/ui/view/dataverse/component/pageTemplate/pageTemplate.tsx
```diff
@@ -19,17 +23,26 @@ const isGeneralMetadata = (metadata: Metadata): metadata is GeneralMetadata =>
...
     A.filter(isTagsMetadata),
     A.chain(metadata => metadata.value)
   )
-
   const generalMetadata = pipe(metadata, A.filter(isGeneralMetadata))
 
   return (
     <div className="okp4-dataverse-portal-dataverse-component-page-template-main">
       <div className="okp4-dataverse-portal-dataverse-page-template-left-side-wrapper">
+        <div className="okp4-dataverse-portal-dataverse-back">
```
ErikssonJoakim, line 40: back doesn't tell us much about the element. Seeing as this a container for the back button I'd call it that

```suggestion
        <div className="okp4-dataverse-portal-dataverse-back-button-container">
```

src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
```diff
@@ -8,5 +10,36 @@
...
+        align-items: center;
+
+        .okp4-dataverse-portal-dataverse-back-button {
+          display: flex;
+          justify-content: center;
+          align-items: center;
+          background: none;
+          border: 0;
+          width: fit-content;
+          padding: 10px 7px;
```
ErikssonJoakim, line 26: This style can be simplified

```suggestion
        display: flex;
        align-items: center;
        column-gap: 5px;

        .okp4-dataverse-portal-dataverse-back-button {
          all: unset;
          display: flex;
          align-items: center;
```

ErikssonJoakim (approved):
Looks good Just have some refactoring suggestions. To do with as you wish 😄

src/ui/view/dataverse/component/pageTemplate/pageTemplate.tsx
0xLodz: good point!

src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
0xLodz: Good point to discuss about. If we define a global style for the button, unsetting all props would give us a different result. In that case, your proposal works.
We can also reset it globally in order to avoid to unset every new button. What do you think @ErikssonJoakim and @lolottetheclash ? Same for other props like `box-sizing`, etc. What are you thoughts about reset? I particularly appreciate this [modern proposal](https://www.joshwcomeau.com/css/custom-css-reset/).

src/ui/i18n/translations/common_de.json
ErikssonJoakim: Ah yes I see I had misunderstood the conversation, sorry. I guess it could serve a purpose if we want to add translations where 'dataverse' wouldn't work

src/ui/view/dataverse/component/pageTemplate/pageTemplate.tsx
ErikssonJoakim: Yes you're right. I would keep it short though.

src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
ErikssonJoakim: Yes we could definitely reset it globally. I think it would be good if we only introduce resets when they are necessary though. 

Seeing as a global reset would implicate all buttons though I would prefer this to be done in a separate refactor PR to better scope the issue

0xLodz: Sure, we don't have to add all of these resets. The goal is to avoid resetting default behaviours in many components. Will make a proposal for it, that can be handled in a different PR

src/ui/view/dataverse/component/pageTemplate/pageTemplate.scss
```diff
@@ -1,7 +1,41 @@
...
+@import '@/ui/style/fonts.scss';
 @import '@/ui/style/mixins.scss';
 
 .okp4-dataverse-portal-dataverse-component-page-template-main {
   @include twelve-column-layout;
+  row-gap: 19px;
+
+  .okp4-dataverse-portal-dataverse-back {
+    @include with-theme {
+      display: flex;
```
ErikssonJoakim, line 11: The positioning of this element in the grid is weak as it implicitly takes the first column in the first row which can give unexpected side effects in the future. 

I would prefer that you give it the whole length of the row.

```suggestion
      @include grid-item($colStart: 1, $colEnd: -1, $rowStart: auto, $rowEnd: auto);
      display: flex;
```

PR #167, opened 2023-04-03
feat/add component skeleton

GeoffreyPicard:
Add Skeleton component 

Capture d’écran 2023-04-03 à 16 39 45

Capture d’écran 2023-04-03 à 16 39 31


0xLodz (changes requested):

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Skeleton: FC<SkeletonProps> = ({ width, height, variant = 'text', fontSize }) => {
+  const classSkeleton = classNames('okp4-dataverse-portal-skeleton', variant)
+  return (
+    <div
+      className={classSkeleton}
+      style={{
+        width: width ? `${width}px` : '100%',
+        height: `${height}px`,
+        fontSize: `${fontSize}px`
```
ccamel, line 20: if no `fontSize` -> `undefinedpx`.

ccamel (changes requested):

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,24 @@
+import type { FC } from 'react'
+import classNames from 'classnames'
+import './skeleton.scss'
+
+type SkeletonProps = {
+  width?: number
+  height?: number
+  fontSize?: number
+  variant?: 'text' | 'rectangular' | 'rounded' | 'circular'
```
lolottetheclash, line 9: I can't see where `rectangular` is defined 🤔

GeoffreyPicard: I will now add rectangular class and let text for thee default value

src/ui/component/skeleton/skeleton.tsx
GeoffreyPicard: React catch it and remove height and fontSize if the value is undefined.
So in CSS there is no height and no font-size
I can cover it with a 100%

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const Skeleton: FC<SkeletonProps> = ({ width, height, variant = 'text', fontSize }) => {
+  const classSkeleton = classNames('okp4-dataverse-portal-skeleton', variant)
+  return (
+    <div
+      className={classSkeleton}
+      style={{
+        width: width ? `${width}px` : '100%',
+        height: `${height}px`,
```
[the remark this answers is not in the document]
GeoffreyPicard: React catch it and remove height and fontSize if the value is undefined.
So in CSS there is no height and no font-size
I can cover it with a 100%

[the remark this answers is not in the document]
0xLodz: Interesting, didn't know that React removed it. That makes sense that you don't need default value, thanks to the non-breakable space set in the pseudo element `content` value.
I would have been more explicit by doing something like`...(height ? {height: ${height}px} : {})`, but it's a bit more verbose and harder to read.

ccamel (approved):

src/ui/component/skeleton/skeleton.scss
```diff
@@ -0,0 +1,44 @@
...
+    &::after {
+      animation: animation-translateX 1.3s linear 0.5s infinite;
+      background: linear-gradient(90deg, transparent, rgba(themed('skeleton-animation'), 0.04), transparent);
+      content: '';
+      position: absolute;
+      transform: translateX(-100%);
+      bottom: 0;
+      left: 0;
+      right: 0;
+      top: 0;
```
lolottetheclash, line 23: You could use the shorthand property `inset: 0 ` which handles top/bottom/left/right 😉

src/ui/component/skeleton/skeleton.scss
```diff
@@ -0,0 +1,44 @@
...
+    overflow: hidden;
+    border-radius: 10px;
+
+    &::before {
+      content: '\00a0';
+    }
+
+    &::after {
+      animation: animation-translateX 1.3s linear 0.5s infinite;
+      background: linear-gradient(90deg, transparent, rgba(themed('skeleton-animation'), 0.04), transparent);
```
lolottetheclash, line 16: Could you add this gradient to the `palette.scss` ?

src/ui/component/skeleton/skeleton.scss
```diff
@@ -0,0 +1,44 @@
...
+      border-radius: 30px;
+    }
+
+    &.circular {
+      border-radius: 50%;
+    }
+  }
+}
+
+@keyframes animation-translateX {
```
lolottetheclash, line 40: Perhaps we can just write `translateX` because `keyframes` is used for animations, so it's seems redundant to include `animation` in the title. What do you think?

src/ui/style/palette.scss
```diff
@@ -62,3 +62,9 @@ $gradient: linear-gradient(270deg, #00b6b7 17.51%, #004073 100%);
 // semantic
 $success: #4ad377;
 $error: #ff3648;
+
+// skeleton
+$skeleton-base-light: darkgrey;
+$skeleton-base-dark: gainsboro;
+$skeleton-animation-light: dimgrey;
```
lolottetheclash, line 69: Could you turn color names into hex code colors to keep consistency with others please?

lolottetheclash (changes requested):

src/ui/component/skeleton/skeleton.scss
0xLodz: Yes it is redundant, we can remove it.

src/ui/component/skeleton/skeleton.scss
0xLodz: this gradient uses a `themed` value, is it still possible to add it to the palette when it depends on a mixin?

src/ui/style/palette.scss
0xLodz: @lolottetheclash  Here is the stylelint error that we have that prevent us to use hex code for these colors:
`Expected "#696969" to be "dimgrey" (color-named)` -> [color-named](https://stylelint.io/user-guide/rules/color-named)
Should we deactivate this rule?

lolottetheclash: ah yes I see that the linter reminds us this rule!
for me we should remove it since we only use hexa, do you agree to remove this rule `"color-named": "always-where-possible"` ? @fredericvilcot ?

0xLodz: I removed the rule and replaced project colors with hexa in [efcc488](https://github.com/okp4/dataverse-portal/pull/167/commits/efcc488681b0f1e35161155419f5e783f1fb9014) and  [939540c](https://github.com/okp4/dataverse-portal/pull/167/commits/939540c7361a020c4fab516df02c8e434f916e67). I can remove these changes if necessary.

src/ui/style/theme.scss
```diff
@@ -63,7 +66,10 @@ $themes: (
     text: $text-dark,
     success: $success,
     error: $error,
-    gradient: $gradient
+    gradient: $gradient,
+    skeleton-base-color: $skeleton-base-color-dark,
+    skeleton-animated-color: $skeleton-animated-color-dark,
+    skeleton-gradient: linear-gradient(90deg, transparent, rgba($skeleton-animated-color-light, 0.04), transparent)
```
lolottetheclash, line 72: Don't you want to use `skeleton-animated-color-dark` here?

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,52 @@
...
+import './skeleton.scss'
+
+type SkeletonProps = {
+  width?: number | string
+  height?: number | string
+  fontSize?: number | string
+  variant?: 'text' | 'rectangular' | 'rounded' | 'circular'
+  className?: string
+  style?: CSSProperties
+  count?: number
```
fredericvilcot, line 12: What's the purpose of those props?
IMHO: 
- `className`: not needed because parent must control the rendering
- `style`: not needed for now, please keep KISS
- `count`: I do not expect this component to map any element.. The purpose of a `Skeleton` component is to render a `Sleleton` component with some style props. I don't want to have any `array` computations if I only want to render one item by one item (e.g a table)

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  style,
+  count = 1
+}) => {
+  const isRounded = variant === 'rounded'
+
+  if (isRounded && !height) {
+    throw new Error(
+      "A height must be provided when using the 'rounded' variant of Skeleton component."
+    )
+  }
```
fredericvilcot, line 30: Great to throw an error, but who will catch it ?

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,52 @@
...
+    ...(height ? { height: typeof height === 'number' ? `${height}px` : height } : {}),
+    ...(fontSize ? { fontSize: typeof fontSize === 'number' ? `${fontSize}px` : fontSize } : {}),
+    ...(borderRadius ? { borderRadius } : {}),
+    ...style
+  }
+
+  const renderSkeletons = (): JSX.Element[] =>
+    Array(count)
+      .fill(null)
+      .map((_, i) => <div className={classSkeleton} key={i} style={mergedStyle}></div>)
```
fredericvilcot, line 49: Please do not render an array, it must be the responsibility of the parent to instantiate an array and map over it.

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return <>{renderSkeletons()}</>
```
fredericvilcot, line 51: Avoid the use of React Fragments, please use HTML `div` tag.

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  variant?: 'text' | 'rectangular' | 'rounded' | 'circular'
+  className?: string
+  style?: CSSProperties
+  count?: number
+}
+
+export const Skeleton: FC<SkeletonProps> = ({
+  width,
+  height,
+  variant = 'text',
```
fredericvilcot, line 18: What's the purpose of the `text` variant and the `fontSize` prop?
Please, stay KISS, just use a rounded variant with a height property to customize the skeleton.

fredericvilcot (changes requested):

src/ui/component/skeleton/skeleton.tsx
0xLodz: @fredericvilcot  `text` variant apply default `height` or `border-radius` properties:
```css
&.text {
  height: 1.2em;
  border-radius: 4px;
 }
```
It's an opinionated variant based on what Material-UI does.
We can make the API more simple by providing a `radius` property, what do you think?
It handles easily all our variant cases, except the rounded one, that must be computed manually.

The `fontSize` is not relevant, we can definitely remove it.

src/ui/component/skeleton/skeleton.tsx
0xLodz: `className` and `style` purpose is to customise the skeleton element.
`count`'s purpose is to avoid to display several similar skeleton in a row.
We can remove all of them.

src/ui/component/skeleton/skeleton.tsx
fredericvilcot: @0xLodz 
I'll suggest to remove the `fontSize` property.
About the `text` variant, IMHO it brings nothing more than a rounded variant with a `height` property and a different radius.
We can keep it for now, but we must keep focusing on small components that do the job for now.

src/ui/component/skeleton/skeleton.tsx
fredericvilcot: I'll suggest to remove the error for now, the `ErrorBoundary` will be added in a future iteration with the 500 page feature.

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  if (isRounded && !height) {
+    console.error(
+      "A height must be provided when using the 'rounded' variant of Skeleton component."
+    )
+  }
+  if (isRounded && height) {
+    const borderRadius =
+      typeof height === 'number' ? `${Math.round(height * 0.5)}px` : `calc(${height} * 0.5)`
+    mergedStyle.borderRadius = borderRadius
+  }
```
lolottetheclash, line 27: I'm not sure you need all these:

```
const isRounded = variant === 'rounded'
  if (isRounded && !height) {
    console.error(
      "A height must be provided when using the 'rounded' variant of Skeleton component."
    )
  }
  if (isRounded && height) {
    const borderRadius =
      typeof height === 'number' ? `${Math.round(height * 0.5)}px` : `calc(${height} * 0.5)`
    mergedStyle.borderRadius = borderRadius
  }
```
  
 I tested without and it works well.
if the height is not specified, it takes the value `auto`, knowing that you have also defined a content with a value `" "` which allows you to have a content to display.
The border radius is defined at 10 px and can stay like that without calculating different border radius.
What do you think?

0xLodz: @lolottetheclash default Skeleton variant is `text`, with an height of 1.2em and a 4px radius.
If we set a wrong variant name like foo, it fallbacks on a 10px radius.

If we remove the computed radius and not set a variant by default, we will end with a 10px radius by default. Is it working for most of use cases? Probably.
If you agree we can remove the code you mentioned.

lolottetheclash: @0xLodz No pb for me to keep the default variant and also the 10px radius, also I don't think that we should have a fallback because the variants possibilities are defined and typed.
I think you can remove the code I copied previously.

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,23 @@
...
+type SkeletonProps = {
+  width?: number | string
+  height?: number | string
+  variant?: 'text' | 'rectangular' | 'rounded' | 'circular'
+}
+
+export const Skeleton: FC<SkeletonProps> = ({ width, height, variant = 'text' }) => {
+  const mergedStyle: CSSProperties = {
+    width: width ? (typeof width === 'number' ? `${width}px` : width) : '100%',
+    ...(height ? { height: typeof height === 'number' ? `${height}px` : height } : {})
```
fredericvilcot, line 14:
```suggestion
    ...(height && { height: typeof height === 'number' ? `${height}px` : height })
```

src/ui/component/skeleton/skeleton.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  return (
+    <div
+      className={classNames('okp4-dataverse-portal-skeleton', variant)}
+      style={mergedStyle}
+    ></div>
+  )
```
fredericvilcot, line 22:
```suggestion
  return (
    <div
      className={classNames('okp4-dataverse-portal-skeleton-main', variant)}
      style={mergedStyle}
     />
  )
```

fredericvilcot (changes requested):

lolottetheclash (approved):
Good for me !

PR #173, opened 2023-04-06
fix(dataverse): position back button

ErikssonJoakim:
Positions the back button on the whole row


lolottetheclash (approved):
Nice catch !

Issue #169, opened 2023-04-05
Feat : Details page - Temporal coverage  


MdechampG:
# Feature: Details page - Temporal coverage  

> **As a** User of the portal on the details page of a dataset,
> **I want** to read the temporal coverage of the dataset 
> **So that** I easily understand the context of this dataset

### Scenario: General display

**Given that** I am a user of the web portal on the detail page of a dataset,
**And** I am using any device,
**When** the `Start date` and `End date` of a dataset are filled out, 
**Then** I should see From `Start date` + To `End date` displayed on 2 lines (mock up 1) 
**When** only the `Start date` of a dataset is filled out, 
**Then** I should see From `Start date` (mock up 2) 
**When** only the `End date` of a dataset is filled out, 
**Then** I should see To `End date` (mock up 3) 
**When** there is no date filled out, 
**Then** I should not see Temporal Coverage

### Scenario : Error

**Given that** I am a user of the web portal on the detail page of a dataset,
**And** I am using any device,
**When** the `Start Date` or the `End date` are not filled out correctly (bad format or `Start Date` later than `End Date` for instance) 
**Then** I should see a dash (mock up 4) 

### Scenario : Internationalization

**Given that** I am a user of the web portal on the detail page of a dataset,
**And** the selected language is English
**Then** the date format should be `MM/dd/yyyy` 
**And** texts should be displayed in English
`From`
`To`

**Given that** I am a user of the web portal on the detail page of a dataset,
**And** the selected language is French
**Then** the date format should be `dd/MM/yyyy` 
**And** texts should be displayed in French
`Depuis le`
`Jusqu'au`

**Given that** I am a user of the web portal on the detail page of a dataset,
**And** the selected language is German
**Then** the date format should be `dd.MM.yyyy` 
**And** texts should be displayed in German
`Von`
`Bis`

### Mockups

1 / ![image.png](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/d5183102-94da-477f-bb2a-99d709674618)

2/ ![image.png](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/164a9cae-e306-4bc1-960d-4f4c11926841)

3/ ![image.png](https://images.zenhubusercontent.com/616e76157d9dbb44e290ed55/5d9281b5-8157-4521-ac45-1fa79d4eb542)

[...]

PR #175, opened 2023-04-11
Feat/temporal coverage

0xLodz:
This PR implements https://github.com/okp4/dataverse-portal/issues/169.

## Example with different `period` metadata values

1. None of them
```ts
{
    category: 'generalMetadata',
    property: 'period',
    value: ['', 'foo']
  }
```

2. Only **start** date
```ts
{
    category: 'generalMetadata',
    property: 'period',
    value: ['2022-12-31T00:00:00+00:00', '']
  }
```

3. Only **end** date
```ts
{
    category: 'generalMetadata',
    property: 'period',
    value: ['', '2022-12-31T00:00:00+00:00']
  }
```

4. **End** date anterior to **start** date
```ts
{
    category: 'generalMetadata',
    property: 'period',
    value: ['2022-12-31T00:00:00+00:00', '2022-01-01T00:00:00+00:00']
  }
```

5. Both **start** and **end** dates

```ts
 {
    category: 'generalMetadata',
    property: 'period',
    value: ['2022-01-01T00:00:00+00:00', '2022-12-31T00:00:00+00:00']
  }
```

image


src/ui/view/dataverse/component/generalMetadata/dateFromTo.tsx
```diff
@@ -0,0 +1,76 @@
...
+  localizedDate,
+  label
+}: {
+  label: 'from' | 'to'
+  localizedDate: string
+}): JSX.Element => {
+  const { t } = useTranslation('common')
+  return (
+    <span>
+      <b>
```
lolottetheclash, line 20: Usage of `<b>` is discouraged now (it was used to render bold text before HTML 4 arrived), it's better to use css for styling.

src/ui/i18n/translations/common_fr.json
```diff
@@ -17,5 +17,8 @@
     "accessGovernance": "Accéder à la gouvernance"
   },
   "filters": "Filtres",
-  "dataverse": "Dataverse"
+  "dataverse": "Dataverse",
+  "from": "Depuis le",
+  "to": "Jusqu'au",
+  "colon": "\u00A0:\u00A0"
```
lolottetheclash, line 23: I'm not sure it's a good practise to deal with unicode character here...

src/ui/page/dataverse/dataset/dataset.tsx
```diff
@@ -9,31 +9,37 @@ import PageTemplate from '@/ui/view/dataverse/component/pageTemplate/pageTemplat
...
     property: 'geographicalCoverage',
     value: 'France'
   },
   {
+    category: 'generalMetadata',
     iconName: 'calendar',
-    property: 'temporalCoverage',
-    value: '2022-01-01T00:00:00+00:00'
+    property: 'period',
+    value: ['2022-12-31T00:00:00+00:00', '2022-01-01T00:00:00+00:00']
```
lolottetheclash, line 39: I'm not sure I understood, I may be wrong!
But here, this is the same exemple than the example n° 4 where you have the start date which is happening after the end date, right? 
Shouldn't it display `-` instead of the from/to date?

lolottetheclash (commented):
Nice job!
Personal thought : I can see that you sometimes choose to name your dates with "from" and "to" and sometimes with "start" and "end" but I think `startDate` and `endDate` are more explicit when we are talking about dates, (but it's up to you) also it's good to keep consistency for namings.

src/ui/page/dataverse/dataset/dataset.tsx
0xLodz: You are right! Thanks a lot, the bug is [in the component](https://github.com/okp4/dataverse-portal/pull/175/files#diff-4d946423f5a1c19c7fe1e8c1410d4d5e72aec0c4e9b9f5ad4c71a886dc0bab1fR46).
The properties have been reversed.

src/ui/i18n/translations/common_fr.json
0xLodz: I understand the concern about using unicode characters, using plain text is more readable. We can replace the Unicode with regular spaces if necessary but this may not provide the same control over spacing and line breaks. 

We can do this: `"colon": " : "`
Or add `&nbsp;` in the component but we would need to do it each time we have to deal with colons.

What do you think?

src/ui/view/dataverse/component/generalMetadata/dateFromTo.tsx
0xLodz: Good point! I will style it with a class.

0xLodz:
> Nice job! Personal thought : I can see that you sometimes choose to name your dates with "from" and "to" and sometimes with "start" and "end" but I think `startDate` and `endDate` are more explicit when we are talking about dates, (but it's up to you) also it's good to keep consistency for namings.

I wanted to distinguish the fact that a start date and end date are not necessary linked, and when used together as a part of a range, `from` and `to` seemed more appropriate. However `from` and `to` can be used in a different context. We can replace them, I do agree that startDate and EndDate are clearer and more explicit.

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type MetadataValueProps = {
+  metadata: ItemGeneralMetadata
+}
+
+export const GeneralMetadataValue = ({ metadata }: MetadataValueProps): JSX.Element | null => {
+  const namespace = 'generalMetadata'
+  const { t, i18n } = useTranslation(namespace)
+  const { value, property } = metadata
+  if (isGeneralMetadata(metadata)) {
```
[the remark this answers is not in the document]
0xLodz: Thanks for pointing this out, didn't included tags as `generalMetadata`. I will update the narrowing functions.

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
```diff
@@ -0,0 +1,51 @@
...
+      value,
+      fromPredicate(
+        () => translationKeyExists,
+        () => value
+      ),
+      E.map(val => t(`${namespace}.${property}.value.${val}`)),
+      E.getOrElse(() => value)
+    )
+    return <span>{translatedValue}</span>
+  }
```
ccamel, line 42: A bit puzzling to me, particularly due to the utilization of "E.fromPredicate" which is surprising. Why not passing in a default value for cases the key could not be found?


```suggestion
  if (isGeneralMetadata(metadata)) {
    return <span>{t(`${namespace}.${property}.value.${value}`, `${value}`)}</span>
  }
```

ccamel (changes requested):

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  if (isPeriodGeneralMetadata(metadata)) {
+    const [startDateValue, endDateValue] = metadata.value
+    return <DateFromTo fromDateString={endDateValue} toDateString={startDateValue} />
+  }
+
+  console.warn('Metadata is not of type ItemGeneralMetadata')
```
ccamel, line 49: I'm not a big fan of logging (is it drop in prod?).
But if you want to log during development to help developers quickly identify potential problems, you should provide more context (context is king).  In this case, the message isn't very helpful. Why not:

```suggestion
  console.warn('Metadata is not of type %s: %O', 'ItemGeneralMetadata', metadata)
```

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
0xLodz: Thanks for pointing this out! Didn't know about the second i18next parameter that acts as a fallback. Usage of `fromPredicate` was not relevant nor necessary in that case.

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
0xLodz: @ccamel  Yes it is currently printed in the console.
We can either:
- display it on in dev based on `NODE_ENV`.
- return a component that display something an explicit error instead of `null`.
We can even do both.
What would you suggest? Thanks

src/ui/view/dataverse/component/generalMetadata/dateFromTo.tsx
0xLodz: @lolottetheclash  On the mockup From and To have the same font weight as the date. I ended up removing the <b> tag.

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
ccamel: AFAI there's a vitejs plugin for that: https://www.npmjs.com/package/vite-plugin-remove-console.
But don't manage it here, create a new ticket for that. We will address it later. Thx :pray: 

Just improve the logs to provide more context.

src/ui/i18n/translations/common_fr.json
lolottetheclash: I think you can manage space directly in translations files, without any unicode. I made it in my actual PR if you want to check !

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
0xLodz: Sounds a better way to handle console logging!

src/ui/i18n/translations/common_fr.json
0xLodz: Done in [c364741](https://github.com/okp4/dataverse-portal/pull/175/commits/c364741e2a5cb183a5112829b0cfcbf9c1ca2d49)!

src/ui/view/dataverse/component/generalMetadata/dateInterval.tsx
```diff
@@ -0,0 +1,83 @@
...
+} from '@/util/isoDateTime/isoDateTime'
+import { activeLanguageWithDefault } from '@/ui/languages/languages'
+
+const FormattedDateSpan = ({
+  localizedDate,
+  label
+}: {
+  label: 'from' | 'to'
+  localizedDate: string
+}): JSX.Element => {
```
fredericvilcot, line 16: If you don't mind, I'll prefer to keep consistency for all the React components by removing the return type and declaring explicitly the `React.FC` type for the function.

src/ui/view/dataverse/component/generalMetadata/dateInterval.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const { t } = useTranslation('common')
+  return (
+    <span>
+      {t(label)}&nbsp;
+      {localizedDate}
+    </span>
+  )
+}
+const InvalidDateFallback = (): JSX.Element => {
```
fredericvilcot, line 25: Same here.

src/ui/view/dataverse/component/generalMetadata/dateInterval.tsx
```diff
@@ -0,0 +1,83 @@
...
+    localizedStartDate,
+    E.fold(
+      () => handleLocalizedEndDate(localizedEndDate),
+      localizedStartDate =>
+        pipe(
+          localizedEndDate,
+          E.fold(
+            () => <FormattedDateSpan label="from" localizedDate={localizedStartDate} />,
+            localizedEndDate =>
+              handleDateRange(startDate, endDate, localizedStartDate, localizedEndDate)
```
fredericvilcot, line 78: It's a little bit weird to me to check for the date range validity at the end of the process after converting the ISO dates to locales.

I would have expected to check first if the dates are valid (e.g. are they valid regarding the conversion and the range?) and that to "fail fast" by returning the `InvalidDateFallback` as soon as possible.

This allows in particular to avoid unnecessary transformations and especially to prevent from giving four parameters to the `handleDateRange` function if the range of dates is not valid from the start.

What do you think?

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const isPeriodGeneralMetadata = (
+  metadata: ItemGeneralMetadata
+): metadata is ItemGeneralMetadata & { property: 'period'; value: [string, string] } =>
+  metadata.property === 'period'
+
+const isGeneralMetadata = (
+  metadata: ItemGeneralMetadata
+): metadata is Omit<ItemGeneralMetadata, 'value'> & { value: string } =>
+  metadata.property !== 'tags'
```
fredericvilcot, line 15: I'm sorry but I still can't understand this type guard.

In [dataspace.tsx](https://github.com/okp4/dataverse-portal/pull/175/files#diff-eb3fbf240f42b523582d0f723848f398521bef291e50fbeb418d2f18fd2bb152), you added the `category` property to the `tags` object and declared its value with `generalMetadata`, that is a good point.

But, here, you say everything that is not `tags` is `GeneralMetadata`.

So, what type of metadata do `tags` belong to?

fredericvilcot (changes requested):

src/ui/view/dataverse/component/generalMetadata/generalMetadataValue.tsx
0xLodz: You are right, `tags` must belong to general metadata.  I updated it in [818e093](https://github.com/okp4/dataverse-portal/pull/175/commits/818e0933963bb263e94957728966ca1687998c86), by reusing the existing narrowing function from the PR #174, that has been recently merged.
Here is the one from #174 that seems correct to me:
```ts
export const isGeneralMetadataWithIcon = (
  metadata: ItemGeneralMetadata
): metadata is Omit<ItemGeneralMetadata, 'value'> & { value: string } =>
  metadata.property !== 'tags' &&
  metadata.category === 'generalMetadata' &&
  isPropertyWithIcon(metadata.property)
  ```

fredericvilcot: @0xLodz 
But, if you say _"tags must belong to general metadata"_, so why the type guard does not include them? (`metadata.property !== 'tags'`)

0xLodz: Gotcha. The `metadata.property !== 'tags'` doesn't make sense here.
I amended the [previous commit](https://github.com/okp4/dataverse-portal/pull/175/commits/9767cbaadd5eb25a0cf867ee042ddb9c04ba0477). Sorry for the misunderstanding.

src/ui/view/dataverse/component/generalMetadata/dateInterval.tsx
0xLodz: Yes it is more simple to read when the date range is validated first. Having as much as parameters in a function is something that I wanted to avoid but didn't think about moving the range validation at the beginning of the sequence. Thanks for the advice, I updated the code accordingly, and created a function to handle both dates.

fredericvilcot (approved):

lolottetheclash (approved):
Good for me!

ccamel (approved):
LGTM! Thanks

PR #182, opened 2023-04-17
Feat/dataverse ontology

fredericvilcot:
ℹ️ This PR adds the `dataverse` core logic domain that allows to retrieve the whole `dataverse` into ontology database.

⚙️This mechanism is based on two principles:

- a `domain` layer that contains the whole elements related to the `dataverse` bounded context. (aggregate)
- an `infrastructure` layer  that allows the domain to communicate with the outside.

🧠 Those two mechanisms work together to: fetch elements in the ontology, map them and compute them to be consumed by any controller.

The controller has access to a public contract interface (store) that provides the base getters (reactive selectors) and setters to perform actions.

🔜 Views must me updated to consume data from the store, regarding the contract specifications.

### DIFF
--- a/src/domain/dataverse/aggregate.ts
+++ b/src/domain/dataverse/aggregate.ts
@@ -1,0 +1,176 @@
+/* eslint-disable max-lines-per-function */
+import { pipe } from 'fp-ts/function'
+import * as E from 'fp-ts/Either'
+import type * as T from 'fp-ts/Task'
+import * as TE from 'fp-ts/TaskEither'
+import { immer } from 'zustand/middleware/immer'
+import { createStore } from 'zustand/vanilla'
+import type { DataverseDTO } from './dto'
+import type { DataverseEntity } from './entity'
+import type { DataversePort } from './port'
+
+type ByTypeQueryFilter = 'all' | 'DataSpace' | 'Dataset' | 'Service'
+
+type DataverseQueryFilters = {
+  byType: ByTypeQueryFilter
+}
+
+type DataverseQueryError = Error | null
+
+type DataverseQuery = {
+  limit: number
+  hasNext: boolean
+  isLoading: boolean
+  error: DataverseQueryError
+  filters: DataverseQueryFilters
+  language: string
+  setIsLoading: (isLoading: boolean) => void
+  setHasNext: (hasNext: boolean) => void
+  setError: (error: DataverseQueryError) => void
+  setFilters: (newFilter: DataverseQueryFilters) => void
+  setLanguage: (newLng: string) => void
+}
+
+export type DataverseAggregate = {
+  dataverse: DataverseEntity
+  query: DataverseQuery
+  addToDataverse: (newDataverse: DataverseEntity) => void
+  clearDataverse: () => void
+  loadDataverse: () => T.Task<void> | void
+}
+
+// Public contract with the controller that consumes the store
+export type DataverseStore = {
+  query: {
+    hasNext: boolean
+    isLoading: boolean
+    error: DataverseQueryError
+    filters: DataverseQueryFilters
+    setFilters: (newFilter: DataverseQueryFilters) => void
+    setLanguage: (newLng: string) => void
+  }
+  dataverse: DataverseDTO
+  loadDataverse: () => T.Task<void> | void
+}
+
+type ImmerSetReturnType = Partial<DataverseAggregate>
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export const dataverseAggregate = (gateway: DataversePort) =>
+  createStore<DataverseAggregate, [['zustand/immer', never]]>(
+    immer<DataverseAggregate>((set, get) => ({
+      dataverse: [],
+      query: {
+        limit: 21,
+        hasNext: false,
+        isLoading: false,
+        error: null,
+        filters: { byType: 'all' },
+        language: 'en',
+        setIsLoading: (isLoading: boolean): void =>
+          set(
+            (state): ImmerSetReturnType => ({
+              query: { ...state.query, isLoading }
+            })
+          ),
+        setHasNext: (hasNext: boolean): void =>
+          set(
+            (state): ImmerSetReturnType => ({
+              query: { ...state.query, hasNext }
+            })
+          ),
+        setError: (error: DataverseQueryError): void =>
+          set(
+            (state): ImmerSetReturnType => ({
+              query: { ...state.query, error }
+            })
+          ),
+        setFilters: (newFilters: DataverseQueryFilters): void => {
+          const { clearDataverse } = get()
+          clearDataverse()
+          return set(
+            (state): ImmerSetReturnType => ({
+              query: { ...state.query, filters: { ...state.query.filters, ...newFilters } }
+            })
+          )
+        },
+        setLanguage: (newLng: string): void => {
+          const {
+            clearDataverse,
+            query: { setError }
+          } = get()
+          if (!newLng.length) {
+            return setError(
+              new Error(
+                `Oops.. An error occurred in <setLanguage> call.. Parameter cannot be empty: <${newLng}>..`
+              )
+            )
+          }
+          clearDataverse()
+          return set(
+            (state): ImmerSetReturnType => ({
+              query: { ...state.query, language: newLng }
+            })
+          )
+        }
+      },
+      clearDataverse: (): void =>
+        set(
+          (): ImmerSetReturnType => ({
+            dataverse: []
+          })
+        ),
+      addToDataverse: (newDataverse: DataverseEntity): void =>
+        set(
+          (state): ImmerSetReturnType => ({
+            dataverse: [...state.dataverse, ...newDataverse]
+          })
+        ),
+      loadDataverse: (): T.Task<void> | void => {
+        const { addToDataverse, query, dataverse } = get()
+        const { setIsLoading, setHasNext, setError, limit, language, filters } = query
+        const offset = dataverse.length
+
+        setIsLoading(true)
+        setError(null)
+
+        const checkLanguageInvariant = (): E.Either<Error, void> =>
+          pipe(
+            E.tryCatch(
+              () => {
+                if (!language.length) {
+                  throw new Error(
+                    'Oops.. An error occurred when trying to load Dataverse.. Language cannot be empty..'
+                  )
+                }
+              },
+              reason =>
+                reason instanceof Error
+                  ? reason
+                  : new Error('Oops.. An unspecified error occurred..')
+            )
+          )
+
+        const fetchDataverse = (): T.Task<void> =>
+          pipe(
+            gateway.retrieveDataverse(language, limit, offset, filters),
+            TE.match(
+              e => {
+                setIsLoading(false)
+                setError(e)
+              },
+              r => {
+                setIsLoading(false)
+                setHasNext(r.query.hasNext)
+                addToDataverse(r.data)
+              }
+            )
+          )
+
+        return pipe(
+          checkLanguageInvariant(),
+          E.matchW(e => setError(e), fetchDataverse)
+        )
+      }
+    }))
+  )
--- a/src/infra/dataverse/sparql/sparqlGateway.ts
+++ b/src/infra/dataverse/sparql/sparqlGateway.ts
@@ -16,7 +16,7 @@ export const sparqlGateway: DataversePort = {
     language: string,
     limit: number,
     offset: number,
-    filters?: RetrieveDataverseQueryFilters
+    filters: RetrieveDataverseQueryFilters
   ): TE.TaskEither<Error, RetrieveDataverseResult> => {
     const query = `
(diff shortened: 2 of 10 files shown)


0xLodz (changes requested):
Great job! It provides a good overview of the domain and ontology DB, and how to store and consume these data. I provided two small feedbacks, thanks.

src/domain/dataverse/aggregate.ts
```diff
@@ -0,0 +1,176 @@
...
+                    'Oops.. An error occurred when trying to load Dataverse.. Language cannot be empty..'
+                  )
+                }
+              },
+              reason =>
+                reason instanceof Error
+                  ? reason
+                  : new Error('Oops.. An unspecified error occurred..')
+            )
+          )
```
ccamel, line 152: Unecessary `pipe`:

```suggestion
            E.tryCatch(
              () => {
                if (!language.length) {
                  throw new Error(
                    'Oops.. An error occurred when trying to load Dataverse.. Language cannot be empty..'
                  )
                }
              },
              reason =>
                reason instanceof Error
                  ? reason
                  : new Error('Oops.. An unspecified error occurred..')
            )

```

src/domain/dataverse/aggregate.ts
```diff
@@ -0,0 +1,176 @@
...
+            gateway.retrieveDataverse(language, limit, offset, filters),
+            TE.match(
+              e => {
+                setIsLoading(false)
+                setError(e)
+              },
+              r => {
+                setIsLoading(false)
+                setHasNext(r.query.hasNext)
+                addToDataverse(r.data)
```
ccamel, line 165: There's a potential race condition here. If `fetchDataverse` is called twice, it's possible that the same result could be inserted more than once. To be confirmed.

The same for `isLoading`.

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+      LIMIT ${limit}
+      OFFSET ${offset}
+`
+
+    const fetchHeaders: RequestInit = {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+        Authorization: `Basic ${APP_ENV.sparql['credentials']}`,
```
ccamel, line 56: Credentials are optional. Remove `Authorization` header if no credentails provided.

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -0,0 +1,142 @@
...
+        accept: 'application/sparql-results+json'
+      },
+      body: `query=${encodeURIComponent(query)}`
+    }
+
+    const fetchDataverse = (): TE.TaskEither<Error, Response> =>
+      TE.tryCatch(
+        async () => fetch(APP_ENV.sparql['endpoint'], fetchHeaders),
+        reason =>
+          reason instanceof Error
```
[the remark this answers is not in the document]
fredericvilcot: @0xLodz 
Thanks for this one, I completely forgot to handle http errors.
I will push the fix.

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -0,0 +1,142 @@
...
+        O.map(a => [b, a])
+      )
+
+    const splitBindingType = ([b, a]: [SparqlBinding, string]): O.Option<
+      [SparqlBinding, string, string]
+    > =>
+      pipe(
+        O.fromNullable(b.type.value.split('/').at(-1)),
+        O.map(c => [b, a, c])
+      )
```
ccamel, line 102: Can be refactored a bit:

```ts
    // this can be put in utils
    const getURILastElement = (str: string): O.Option<string> =>
      O.fromNullable(str.split('/').at(-1))

    const splitBindingId = (b: SparqlBinding): O.Option<[SparqlBinding, string]> =>
      pipe(
        b.id.value,
        getURILastElement,
        O.map(a => [b, a])
      )

    const splitBindingType = ([b, a]: [SparqlBinding, string]): O.Option<
      [SparqlBinding, string, string]
    > =>
      pipe(
        b.type.value,
        getURILastElement,
        O.map(c => [b, a, c])
      )
```

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -0,0 +1,142 @@
...
+        'Content-Type': 'application/x-www-form-urlencoded',
+        Authorization: `Basic ${APP_ENV.sparql['credentials']}`,
+        accept: 'application/sparql-results+json'
+      },
+      body: `query=${encodeURIComponent(query)}`
+    }
+
+    const fetchDataverse = (): TE.TaskEither<Error, Response> =>
+      TE.tryCatch(
+        async () => fetch(APP_ENV.sparql['endpoint'], fetchHeaders),
```
ccamel: Missing HTTP error code management here. 😅

ccamel (commented):
Good work. I'll leave some comments to your judgment. 😌

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -0,0 +1,142 @@
...
+    )
+    const filterDataverseEntity = (dataverseEntity: DataverseEntity): DataverseEntity =>
+      pipe(
+        dataverseEntity,
+        getDataverseEntityLengthOption,
+        O.match(
+          () => dataverseEntity,
+          e => A.dropRight(1)(e)
+        )
+      )
```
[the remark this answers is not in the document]
fredericvilcot: I want to remove the extra element of the entity, to store `limit - 1` elements in the array.
Here `limit` equals `21`, so I'll suggest to do this instead: 

```ts
const filterDataverseEntity = A.takeLeft(limit - 1)
```

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: Yes, I committed the fix [here](https://github.com/okp4/dataverse-portal/pull/182/commits/387001d35bc4249fa381ef86138ee3465a2d22dc)

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: As ontology DB is now opened with a read-only access for anonymous, I will remove `Authorization ` header as well as `env` variables.

src/domain/dataverse/aggregate.ts
fredericvilcot: Yep, I'll remove it.

src/infra/dataverse/sparql/sparqlGateway.ts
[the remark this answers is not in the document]
ccamel: In fact, the limit should represent the maximum number of elements retrieved. I believe it would be more logical to set the limit to 20, for example, and then call the gateway with limit + 1 (21). This would create a more consistent approach.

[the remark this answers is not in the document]
fredericvilcot: Make sense, I'll make the change.

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: Yep, much more clearer.

src/domain/dataverse/aggregate.ts
```diff
@@ -0,0 +1,176 @@
...
+  setFilters: (newFilter: DataverseQueryFilters) => void
+  setLanguage: (newLng: string) => void
+}
+
+export type DataverseAggregate = {
+  dataverse: DataverseEntity
+  query: DataverseQuery
+  addToDataverse: (newDataverse: DataverseEntity) => void
+  clearDataverse: () => void
+  loadDataverse: () => T.Task<void> | void
```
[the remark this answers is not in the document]
fredericvilcot: @ccamel 
I can't return only a `T.Task<void>` because the language invariant method `checkLanguageInvariant` returns an `Option<Error>` that is consumed thanks to pattern matching like that: 

```ts
      loadDataverse: (): T.Task<void> | void => {
        const { addToDataverse, query, dataverse } = get()
        const { setIsLoading, setHasNext, setError, limit, language, filters } = query
        const offset = dataverse.length

        setIsLoading(true)
        setError(O.none)

        const checkLanguageInvariant = (): O.Option<Error> =>
          !language.length
            ? O.some(
                new Error(
                  'Oops.. An error occurred when trying to load Dataverse.. Language cannot be empty..'
                )
              )
            : O.none

        const fetchDataverse = (): T.Task<void> =>
          pipe(
            gateway.retrieveDataverse(language, limit, offset, filters),
            TE.match(
              e => {
                setIsLoading(false)
                setError(O.some(e))
              },
              r => {
                setIsLoading(false)
                setHasNext(r.query.hasNext)
                addToDataverse(r.data)
              }
            )
          )

        return pipe(
          checkLanguageInvariant(),
          O.matchW(fetchDataverse, e => setError(O.some(e)))
        )
      }
```

How can we chain the `checkLanguageInvariant` and the `fetchDataverse` to return a `Task` in both cases?

src/domain/dataverse/aggregate.ts
```diff
@@ -0,0 +1,176 @@
...
+        ),
+      loadDataverse: (): T.Task<void> | void => {
+        const { addToDataverse, query, dataverse } = get()
+        const { setIsLoading, setHasNext, setError, limit, language, filters } = query
+        const offset = dataverse.length
+
+        setIsLoading(true)
+        setError(null)
+
+        const checkLanguageInvariant = (): E.Either<Error, void> =>
```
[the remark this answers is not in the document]
fredericvilcot: Yes, much more simpler, I'll push the change.

src/domain/dataverse/aggregate.ts
```diff
@@ -0,0 +1,176 @@
...
+            dataverse: [...state.dataverse, ...newDataverse]
+          })
+        ),
+      loadDataverse: (): T.Task<void> | void => {
+        const { addToDataverse, query, dataverse } = get()
+        const { setIsLoading, setHasNext, setError, limit, language, filters } = query
+        const offset = dataverse.length
+
+        setIsLoading(true)
+        setError(null)
```
ccamel, line 135: I didn't catched that. You must avoid side effects in this function, as it is pure. Instead, you should place these impure statements in the `Task` monad.

src/domain/dataverse/aggregate.ts
ccamel: 1. will not solve the race condition, I presume. If you capture the condition in the asynchronous function, then you will have a risk of overwriting with 3 (or more) concurrent calls. Moreover, changing the interface contract makes the caller responsible for deciding which function to call in which case, when it should not.
2. `isLoading` is your flag since it's part of the state. Then check & eject.
3. use a simple lock mechanism. Then check & stack.

up to you.

src/domain/dataverse/aggregate.ts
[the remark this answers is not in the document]
ccamel: your `matchW` is not correct. Use a plain `match` instead.

Something like:

```ts
        return pipe(
          checkLanguageInvariant(),
          O.match(fetchDataverse, e => { setError(O.some(e)); return T.of(undefined) })
        )
```

or

```ts
        return pipe(
          checkLanguageInvariant(), 
          O.match(fetchDataverse, flow(O.some, setError, T.of)))
```

or 

...

src/domain/dataverse/aggregate.ts
fredericvilcot: @ccamel 
I'm not sure to understand, we are already here in a TE monad, using matching pattern to execute side effects.
What do you mean by _ you should place these impure statements in the Task monad._ ?

ccamel: @fredericvilcot 

It's mainly the question of effects. If you call your function, it shouldn't produce any effect, it should just return the value `TaskEither`. This is not the case in your function, because if you call the function, you set the state to isLoading to true. The side effect should be inside the TE.

```ts
loadDataverse: (): T.Task<void> | void => {
        const { addToDataverse, query, dataverse } = get()
        const { setIsLoading, setHasNext, setError, limit, language, filters } = query
        const offset = dataverse.length

        setIsLoading(true) // <- side effect outside the TE
        setError(null) // <- side effect outside the TE

        return TE...
```

fredericvilcot: @ccamel 
Sorry, I thought you were speaking about the `set...` functions inside the `TE.match` function.
I didn't noticed you were referring to these lines...
100% agree, I will wrap these side effects into a function that's called inside the `pipe`.

fredericvilcot:
@ccamel @0xLodz 
I made the changes, pls have a review

lolottetheclash (approved):
You made a great Job !!! 👌💪🚀

0xLodz (approved):
Well done!

ccamel (approved):
Great! Let's go! :rocket:

Issue #179, opened 2023-04-13
Chore: Remove console logs in production environment


0xLodz:
# Chore: Remove console logs in production environment

> **As a** developer,
> **I want** to see console logs only non-production environments
> **So that** no console logs appear in the production environment

### Scenario

**Given that** I am a user of the Dataverse Portal,
**And** I am using a *desktop* device,
**When** I open the browser developer tools in production
**Then** I don't see any console.log from the application.
**And** I only see console.log from the application in non-production environments

## Requirements

1. Add a Vite plugin or any similar solution that help us to deal with console logs without adding conditions in the code based on the env.

## Resources

- [discussion about the requirement](https://github.com/okp4/dataverse-portal/pull/175#discussion_r1165290775)

- [Vite plugin](https://www.npmjs.com/package/vite-plugin-remove-console)

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206528851) by [Unito](https://www.unito.io)

ChaimaaZaoui:
I can't regenerate the case with my accounts, @0xLodz tested and it's all ok

PR #186, opened 2023-04-20
chore: prevent production logs

0xLodz:
This PR implements https://github.com/okp4/dataverse-portal/issues/179.

### How to test it locally?
From this branch `chore/prevent-production-logs`,
- add some console.logs whenever you want and test them locally with `npm run dev` (port 5173).
- build your app with `npm run build`, then launch and test it with `npm run preview` (port 4173).

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -8,23 +8,19 @@ import htmlConfigOptions from './config/html-config'
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [
-    htmlConfig(htmlConfigOptions),
-    react(), 
-    svgr()
-  ],
+  plugins: [htmlConfig(htmlConfigOptions), react(), svgr()],
   resolve: {
     alias: {
       '@': path.resolve(__dirname, './src')
     }
   },
   build: {
-      rollupOptions: {
-        output: {
-          manualChunks: {
-            'react-vendor': ['react', 'react-dom', 'react-router-dom']
-          }
+    rollupOptions: {
+      output: {
+        manualChunks: {
+          'react-vendor': ['react', 'react-dom', 'react-router-dom']
         }
       }
-    },
+    }
+  }
 })
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -7,12 +7,15 @@ import htmlConfig from 'vite-plugin-html-config'
 import htmlConfigOptions from './config/html-config'
 
 // https://vitejs.dev/config/
-export default defineConfig({
+export default defineConfig(({ mode }) => ({
   plugins: [htmlConfig(htmlConfigOptions), react(), svgr()],
   resolve: {
     alias: {
       '@': path.resolve(__dirname, './src')
     }
+  },
+  esbuild: {
+    drop: mode === 'production' ? ['console', 'debugger'] : []
   },
   build: {
     rollupOptions: {
@@ -23,4 +26,4 @@ export default defineConfig({
       }
     }
   }
-})
+}))


lolottetheclash:
I wonder if we could do without a new plugin and use the quick config instead 🤔
We could add this: 

```
esbuild: {
    drop: ['console', 'debugger']
  }
```

Knowing that even though Vitejs uses Rollup for the bundle, Esbuild is still used for transpilation and minification in Vite during build. I'm not sure if it's a good idea but maybe we could consider passing this config, what do you think @0xLodz 
 & @fredericvilcot ?

0xLodz:
> I wonder if we could do without a new plugin and use the quick config instead 🤔 We could add this:
> 
> ```
> esbuild: {
>     drop: ['console', 'debugger']
>   }
> ```
> 
> Knowing that even though Vitejs uses Rollup for the bundle, Esbuild is still used for transpilation and minification in Vite during build. I'm not sure if it's a good idea but maybe we could consider passing this config, what do you think @0xLodz & @fredericvilcot ?

@lolottetheclash thanks, wasn't aware of this built-in config. There are more details [here](https://github.com/vitejs/vite/discussions/7920#discussioncomment-2709119). Sounds better alternative.

fredericvilcot:
> I wonder if we could do without a new plugin and use the quick config instead 🤔 We could add this:
> 
> ```
> esbuild: {
>     drop: ['console', 'debugger']
>   }
> ```
> 
> Knowing that even though Vitejs uses Rollup for the bundle, Esbuild is still used for transpilation and minification in Vite during build. I'm not sure if it's a good idea but maybe we could consider passing this config, what do you think @0xLodz & @fredericvilcot ?

Yes, sounds to be a recommended practice by the Vite's team.
I would suggest however to discriminate by the environment to keep logs on development mode 😌

lolottetheclash:
> > I wonder if we could do without a new plugin and use the quick config instead 🤔 We could add this:
> > ```
> > esbuild: {
> >     drop: ['console', 'debugger']
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Knowing that even though Vitejs uses Rollup for the bundle, Esbuild is still used for transpilation and minification in Vite during build. I'm not sure if it's a good idea but maybe we could consider passing this config, what do you think @0xLodz & @fredericvilcot ?
> 
> Yes, sounds to be a recommended practice by the Vite's team. I would suggest however to discriminate by the environment to keep logs on development mode 😌

I am not sure to understand... logs are still working on dev mode, am I missing something?

fredericvilcot:
> > > I wonder if we could do without a new plugin and use the quick config instead 🤔 We could add this:
> > > ```
> > > esbuild: {
> > >     drop: ['console', 'debugger']
> > >   }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Knowing that even though Vitejs uses Rollup for the bundle, Esbuild is still used for transpilation and minification in Vite during build. I'm not sure if it's a good idea but maybe we could consider passing this config, what do you think @0xLodz & @fredericvilcot ?
> > 
> > 
> > Yes, sounds to be a recommended practice by the Vite's team. I would suggest however to discriminate by the environment to keep logs on development mode 😌
> 
> I am not sure to understand... logs are still working on dev mode, am I missing something?

You may want to keep logs in development mode, to facilitate debugging.
So, as far as I see, we must discriminate env in config like in this example: https://github.com/vitejs/vite/discussions/7920#discussioncomment-4796266

lolottetheclash:
> > > > I wonder if we could do without a new plugin and use the quick config instead 🤔 We could add this:
> > > > ```
> > > > esbuild: {
> > > >     drop: ['console', 'debugger']
> > > >   }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Knowing that even though Vitejs uses Rollup for the bundle, Esbuild is still used for transpilation and minification in Vite during build. I'm not sure if it's a good idea but maybe we could consider passing this config, what do you think @0xLodz & @fredericvilcot ?
> > > 
> > > 
> > > Yes, sounds to be a recommended practice by the Vite's team. I would suggest however to discriminate by the environment to keep logs on development mode 😌
> > 
> > 
> > I am not sure to understand... logs are still working on dev mode, am I missing something?
> 
> You may want to keep logs in development mode, to facilitate debugging. So, as far as I see, we must discriminate env in config like in this example: [vitejs/vite#7920 (reply in thread)](https://github.com/vitejs/vite/discussions/7920#discussioncomment-4796266)

Ok I see, but my question is: transpiling and minification are not only happening when building? Is it mandatory to discriminate environments here?

fredericvilcot:
> 

Yes, but when it comes to run the dev server, there is also a build ;)

0xLodz:
@lolottetheclash @fredericvilcot the config is now updated using the built-in eslint minification drop option.

fredericvilcot (approved):
Great, thx ☺️

Issue #172, opened 2023-04-06
[Error] - Server problem


ChaimaaZaoui:
# Feature: Server problem
> **As an** end user navigating the portal ,
> **I want** to know when a server problem is encountered
> **So that** I can reacte accordingly to solve the problem

### Scenario I : Default display (desktop / Tablet)
**Given that** I am an end user navigating the web portal,
**When** I click on a button or link 
**And** a server problem is encountered
**Then** I should receive an error snackbar while staying on my page
**And** a text is displayed on the snackbar : 
> Sorry, something went wrong.
> we are currently unable to handle your request.
> Try to refresh this page.

**And** a button is available on the right hand side of the bar : 

* `Refresh`

### Scenario II : Mobile display 
**Given that** I am an end user navigating the web portal,
**When** I click on a button or link 
**And** a server problem is encountered
**Then** I should receive an error snackbar while staying on my page
**And** a text is displayed on the snackbar : 
> Sorry, something went wrong.
> we are currently unable to handle your request.
> Try to refresh this page.

**And** a button is available below the text : 

* `Refresh`

### Scenario III : Refresh button 
**Given that** I am an end user who received a server error snackbar,
**When** I click on the `Refresh` button
**Then** the browser refreshes the page I am currently in

### Scenario IV : Exit snack bar - Desktop
**Given that** I am an end user who received a server error snackbar
**And** I don't want to refresh the page
**And** I want to close the snackbar 
**When** press the `Exit` button on my keyboard
**Then** the snackbar disappears and I am on the initial page where the problem occurred

### Scenario V : Exit snack bar - Mobile / tablet
**Given that** I am an end user who received a server error snackbar
**And** I don't want to refresh the page
**And** I want to close the snackbar 
**When** click anywhere on the screen outside the snackbar
**Then** the snackbar disappears and I am on the initial page where the problem occurred

### Scenario VI: Internalization
Following the selected language, the content on the page is to be displayed accordingly : 

**ENG**
> Sorry, a problem has occurred.
> We are currently unable to process your request.
> Please try refreshing this page.
> `Refresh`

**FR**
> Désolé, un problème est survenu.
> Nous sommes actuellement incapable de traiter votre demande.
> Essayez de rafraîchir cette page.
> `Actualiser`

**DE**
> Es ist leider ein Problem aufgetreten.
> Wir können Ihre Anfrage derzeit nicht bearbeiten.
> Bitte versuchen Sie, diese Seite zu aktualisieren.
[...]

PR #195, opened 2023-04-24
Feat/notifications errors

loiclaudet:
This PR implements https://github.com/okp4/dataverse-portal/issues/172, which contains technical specifications related to this feature.

## Usage
1. Import the notification hook:
```tsx
const dispatchNotification = useDispatchNotification()
```
2. Dispatch a notification from a callback or a react hook (effect, ...)
```tsx
dispatchNotification({
  titleKey: 'error.problem',
  messageKey: 'error.processing',
  type: 'warning',
  action: 'refresh'
})
```
**Note**: We pass title and message translations keys as parameters to the `dispatchNotification` function in order to simplify the usage and gain readability. We thus avoid multiple imports like translation files, `useTranslation` hook, `short-uuid` lib to create the id. It comes with a trade-off that we must pass translated messages keys, an no plain text.

## Toast display for each breakpoints

**EDIT**: Toasts w/o actions are not stacked. New one are displayed above the previous ones. Technical spec and screenshot were updated accordingly.

### Mobile
image

### Tablet
image

### Desktop
image

### Large desktop
image

## Light theme
image

### DIFF
--- a/src/ui/component/notification/notification.tsx
+++ b/src/ui/component/notification/notification.tsx
@@ -16,7 +16,6 @@ export const Notification: FC = () => {
     }))
   )
   const theme = useAppStore(store => store.theme)
-  const notificationData = useMemo(() => notifications(), [notifications])
   const notificationIconByType = useMemo(
     () =>
       new Map<NotificationType, IconName>([
@@ -36,7 +35,7 @@ export const Notification: FC = () => {
 
   return (
     <div className="okp4-dataverse-portal-notifications">
-      {notificationData.map(({ id, type, title, message, action }) => (
+      {notifications().map(({ id, type, title, message, action }) => (
         <Toast
           action={action}
           duration={action ? Infinity : 3000}


src/domain/notification/aggregate.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export type NotificationDTO<T> = {
+  id: NotificationID
+  type: NotificationType
+  title: string
+  message?: string
+  action?: T
+}
+
+export type NotificationsDTO<T> = NotificationDTO<T>[]
```
fredericvilcot, line 45: Maybe it would have been more readable to move DTO types into a dedicated `dto.ts` file

src/domain/notification/aggregate.ts
```diff
@@ -0,0 +1,115 @@
...
+
+type NotificationQuery<T> = {
+  notifications: () => IO<NotificationsDTO<T>>
+}
+
+type NotificationStore<T> = NotificationState<T> & NotificationAction<T> & NotificationQuery<T>
+export type NotificationAggregate<T> = ForgetType<NotificationState<T>, NotificationStore<T>>
```
fredericvilcot, line 52: I would have reversed the names, because, IMHO, the `notificationAggregate` object's type should rather be a `NotificationAggregate`.

src/ui/hook/useDispatchNotification.ts
```diff
@@ -0,0 +1,57 @@
+import { useCallback } from 'react'
+import { useTranslation } from 'react-i18next'
+import { shallow } from 'zustand/shallow'
+import uuid from 'short-uuid'
+import type { ActionType } from '@/ui/store'
+import { useNotificationStore } from '@/ui/store'
+import type { NotificationType } from '@/domain/notification/entity'
```
fredericvilcot, line 7: I would have not imported this type straight from `entity` types to not leak them.

src/ui/component/notifications/notifications.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { useNotificationStore } from '@/ui/store'
+import type { DismissNotificationInput } from '@/domain/notification/aggregate'
+
+export const Notifications: FC = () => {
+  const { notifications, dismissNotification } = toEffectfulObject(
+    useNotificationStore(state => ({
+      notifications: state.notifications,
+      dismissNotification: state.dismissNotification
+    }))
+  )
```
fredericvilcot, line 17: I would have enhanced the `useDispatchNotification` hook to return the whole store abstraction:

```ts
...
return [notifications, {dispatch, dismiss}]

```

src/ui/component/notifications/notifications.tsx
fredericvilcot, line 1: This should be a singular folder name (e.g. `src/ui/component/notification/notification.tsx`)

src/ui/component/notifications/notifications.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const handleClose = useCallback(
+    (notificationInput: DismissNotificationInput) => {
+      dismissNotification(notificationInput)
+    },
+    [dismissNotification]
+  )
+
+  return (
+    <div className="okp4-dataverse-portal-notifications">
+      {notifications().map(({ id, type, title, message, action }, index) => (
```
fredericvilcot, line 38: Could be optimized with a `useMemo` usage to avoid unnecessary computations in rendering

src/ui/component/toast/toast.tsx
```diff
@@ -1,51 +1,140 @@
...
+import { useOnKeyboard } from '@/ui/hook/useOnKeyboard'
+import { Button } from '@/ui/component//button/button'
+import type { DismissNotificationInput } from '@/domain/notification/aggregate'
+import '@/ui/component/notifications/i18n'
 import './toast.scss'
 
-type ToastVariants = 'success' | 'error' | 'warning' | 'info'
+type ToastCTAProps = {
+  type: NotificationType
+  action: string
```
fredericvilcot, line 17: Should be typed with the ActionType enum

src/ui/component/toast/toast.tsx
```diff
@@ -1,51 +1,140 @@
...
+    window.location.reload()
+  }, [])
+  return (
+    <Button
+      className={classNames('okp4-dataverse-portal-toast-cta', type)}
+      label={t(`action.${action}`)}
+      onClick={handleClick}
+    />
+  )
+}
```
fredericvilcot, line 32: This component looks not extensible.

For now, it only handles refresh actions, but how will you do to manage more actions in a next future if you hard code `window.location.reload()` in  `handleClick` function?

src/ui/component/toast/toast.tsx
```diff
@@ -1,51 +1,140 @@
...
   onClose,
-  open = false,
-  variant = 'info'
-}: ToastProps): JSX.Element | null => {
+  message,
+  iconName,
+  action,
+  index
+}) => {
+  const [isHidding, setIsHidding] = useState(false)
```
fredericvilcot, line 56: Could be more explicit naming...

src/ui/component/toast/toast.tsx
```diff
@@ -1,51 +1,140 @@
...
+  const autoHideDuration = useMemo((): number => {
+    const messagesLength = title.length + (message?.length ?? 0)
+    const minDuration = 2000
+    const maxDuration = 7000
+    const durationPerChar = 50
+    const duration = messagesLength * durationPerChar
+    return Math.min(Math.max(duration, minDuration), maxDuration)
+  }, [title, message])
+
+  const handleAction = useCallback(() => {
```
fredericvilcot, line 72: I will not have named this function `handleAction` because this is a bit confusing with the `action` property..

src/ui/component/toast/toast.tsx
```diff
@@ -1,51 +1,140 @@
...
+
   useEffect(() => {
-    if (open && autoHideDuration) {
-      const autoHideDurationTimer = setTimeout(() => {
-        onClose?.()
-      }, autoHideDuration)
-      return () => {
-        clearTimeout(autoHideDurationTimer)
+    if (!isHidding) return
+    timeoutRef.current = setTimeout(() => {
```
fredericvilcot, line 87: I'm not sure you need to memoize the `setTimeout` return's value in a `React.Ref`.

The first version using a `constant` looked good to me and seems more straightforward , what do you think ?

src/ui/component/toast/toast.tsx
```diff
@@ -1,51 +1,140 @@
...
+  const stackedPosition = useRef<number>(index + 1)
+
+  const autoHideDuration = useMemo((): number => {
+    const messagesLength = title.length + (message?.length ?? 0)
+    const minDuration = 2000
+    const maxDuration = 7000
+    const durationPerChar = 50
+    const duration = messagesLength * durationPerChar
+    return Math.min(Math.max(duration, minDuration), maxDuration)
+  }, [title, message])
```
fredericvilcot, line 70: Have all these values been approved by the product team?

fredericvilcot (changes requested):
Great piece of work, that's a good way to initiate the notifications hub in the domain logic!

I wrote down some remarks we can discuss together 😌

src/domain/notification/aggregate.ts
```diff
@@ -0,0 +1,115 @@
...
+import type { Reader } from 'fp-ts/lib/Reader'
+import * as A from 'fp-ts/lib/Array'
+import * as O from 'fp-ts/lib/Option'
+import { pipe } from 'fp-ts/lib/function'
+import type { ForgetType } from '@/util/type'
+import { eqNotification } from './entity'
+import type { NotificationID, NotificationType, Notifications } from './entity'
+
+type NotificationState<T> = {
+  aggregate: Notifications<T>
```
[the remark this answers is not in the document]
ccamel: Yes, I'm not happy with this naming too. 

Well, to my eyes it's more the aggregate roots:

```suggestion
  aggregateRoots: Notifications<T>
```

a pointer to the list of all notifications which are actually aggregate roots.

src/domain/notification/aggregate.ts
ccamel: IMHO, it's best to avoid the use of DTOs in the long term. A request/response approach specific to APIs should be adopted, as is the case here, distinguishing between what is `input` and what is `output`, with different types for each interaction.

src/domain/notification/aggregate.ts
ccamel: Well, the naming is not great.
I think that to avoid any confusion, we should name it `Domain`, which is the visible part of what a domain is, and which allows to interact with an agregate (or several aggregates).

```suggestion
type NotificationStore<T> = NotificationState<T> & NotificationAction<T> & NotificationQuery<T>
export type NotificationDomain<T> = ForgetType<NotificationState<T>, NotificationStore<T>>
```

src/domain/notification/aggregate.ts
```diff
@@ -0,0 +1,115 @@
...
+
+type NotificationQuery<T> = {
+  notifications: () => IO<NotificationsDTO<T>>
+}
+
+type NotificationStore<T> = NotificationState<T> & NotificationAction<T> & NotificationQuery<T>
+export type NotificationAggregate<T> = ForgetType<NotificationState<T>, NotificationStore<T>>
+
+export type NotificationOptions<T> = {
+  initialState: NotificationState<T>
```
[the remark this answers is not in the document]
ccamel: Test suite, indeed.

src/ui/component/toast/toast.tsx
```diff
@@ -1,51 +1,140 @@
...
+  const containerRef = useRef<HTMLDivElement | null>(null)
+  const timeoutRef = useRef<ReturnType<typeof setTimeout>>()
+  const autoHideTimeoutRef = useRef<ReturnType<typeof setTimeout>>()
+  const animationDurationRef = useRef<number>(500)
+  const stackedPosition = useRef<number>(index + 1)
+
+  const autoHideDuration = useMemo((): number => {
+    const messagesLength = title.length + (message?.length ?? 0)
+    const minDuration = 2000
+    const maxDuration = 7000
```
ccamel, line 66: Make this configurable. Everything exists to do this easily.

src/ui/hook/useDispatchNotification.ts
ccamel: That's a good point. But it's not an entity but a Value Object. In general, value objects in DDD don't need to have a mapping to DTOs, mainly because they are typically immutable and  don't have an identity of their own. But this is an open question.

src/ui/store/index.ts
```diff
@@ -1,6 +1,16 @@
...
-import type { StoreApi } from 'zustand'
 
 export const dataverseStore = dataverseAggregate(sparqlGateway) as StoreApi<DataverseStore>
+
+export type ActionType = 'refresh' // TODO: to be completed & moved to the proper place
+const notificationStore = notificationAggregate<ActionType>()()
+export const useNotificationStore = <T>(
+  selector: (state: NotificationAggregate<ActionType>) => T,
+  equals?: (a: T, b: T) => boolean
+): T => useStore(notificationStore, selector, equals)
```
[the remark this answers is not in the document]
ccamel: It's working because `notificationStore` is actually a constant, and `useStore` a hook.

src/domain/notification/aggregate.ts
ccamel: and rename the file `aggregate.ts` to `domain.ts`?

fredericvilcot: @ccamel 
Yes, I agree.
`aggregate` looks too much confusing to my eyes.

fredericvilcot: Another approach linked to DDD structures could consist in naming the file with its purpose, like `notification.ts`

src/ui/hook/useDispatchNotification.ts
fredericvilcot: Yes, you are right, the `NotificationType` is not an entity itself, but it is part of the file that holds the `Notification` entity and I think it's not a good practice to directly import it if from a controller because there is no control on it, IMHO.

src/domain/notification/aggregate.ts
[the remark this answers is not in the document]
fredericvilcot: Yes, but when I come to read about DDD structures, the most part of time, the root object in an aggregate is an entity: 

_"An Aggregate is a set of business objects (Value Objects or Entities) linked together. Among these objects, only one (in general, an Entity) will have a particular role: the role of root of the Aggregate. Any changes made to the aggregate by the rest of the code will have to go through a root method. It is therefore forbidden to modify the aggregate by directly modifying an unrooted object."_

So the notion of `aggregate...` looks a bit confusing for me for the naming of this entity...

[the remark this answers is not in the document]
ccamel: I don't clearly understand the point. `Notifications` is the agregate root (but not an entity), since it maintains the link between all the objects, which are entities (`Notification`).

We could have define the aggregate root that way:

```ts
export type Notifications<T> = Notification<T>[] & AggregateRoot
```

and the `State` is actually a "repository" - a place where all the aggregate roots are kept.

If it is confusing, we can simply name it "data":

```suggestion
  data: Notifications<T>
```

package.json
```diff
@@ -24,17 +24,20 @@
...
-    "react-router-dom": "^6.8.1"
+    "react-router-dom": "^6.8.1",
+    "short-uuid": "^4.2.2"
   },
   "devDependencies": {
     "@okp4/eslint-config": "^1.2.0",
     "@radix-ui/react-switch": "^1.0.1",
     "@swc/jest": "^0.2.24",
     "@types/jest": "^29.4.0",
+    "@types/node": "^18.16.2",
```
ccamel, line 40: Remove this one.

src/domain/notification/aggregate.ts
```diff
@@ -0,0 +1,115 @@
...
+                type: it.type,
+                title: it.title,
+                message: O.toUndefined(it.message),
+                action: O.toUndefined(it.action)
+              }))
+            )
+        })),
+        {
+          anonymousActionType: 'notificationAggregate',
+          enabled: process.env.NODE_ENV === 'development'
```
ccamel, line 112: See: https://vitejs.dev/guide/env-and-mode.html

```suggestion
          enabled: import.meta.env.DEV
```

ccamel (changes requested):

src/domain/notification/aggregate.ts
0xLodz: Neat! Thank you

src/ui/component/toast/toast.tsx
0xLodz: No, I removed it in [6a30ad9](https://github.com/okp4/dataverse-portal/pull/195/commits/6a30ad9d5a347ed926a5c3e4d02a13788388b5fb).

src/ui/component/toast/toast.tsx
0xLodz: We don't need to you're right. Thanks!

src/ui/component/toast/toast.tsx
0xLodz: Good point! I replaced in with `onClose` in [6a30ad9](https://github.com/okp4/dataverse-portal/pull/195/commits/6a30ad9d5a347ed926a5c3e4d02a13788388b5fb)

src/ui/component/toast/toast.tsx
0xLodz: Yes! I opted for `isOpen` state, which is more common in such component

src/ui/component/toast/toast.tsx
0xLodz: As is it's not extensible. We might define an object with key props as `ActionType`, and value would be the function that run side effects. What do you think?

src/ui/component/notifications/notifications.tsx
0xLodz: Good suggestion. Enhancement has been made in [4b19bdf](https://github.com/okp4/dataverse-portal/pull/195/commits/4b19bdf7d81e3f147407fbf553ede674d564c970).

The implementation was slightly different to your proposal in order to avoid such import:
```tsx
const [, { dispatchNotification }] = useDispatchNotification()
```
or
```tsx
const [{ dispatchNotification }] = useDispatchNotification()
```



With the implementation we now have :
```tsx
const { dispatchNotification } = useDispatchNotification()
```

Is it fine?

src/domain/notification/aggregate.ts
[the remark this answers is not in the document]
ccamel: Sure @fredericvilcot. What I propose: I made some organizational evolutions on the PR #213 that is more advanced than this one. I propose a session of pair programming to act together a canonical model of domains that we will then apply on this PR, if it suits you.

FYI @lolottetheclash

src/ui/hook/useDispatchNotification.ts
ccamel: Okay, that makes sense. So let's adopt a strict separation of types between layers.

src/domain/notification/aggregate.ts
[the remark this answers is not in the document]
0xLodz: @ccamel @fredericvilcot do we apply these updates in a next iteration?

src/ui/component/notifications/notifications.tsx
0xLodz: Done in 182c1d6b39e9826eae731ee69ea22c5b0ed84cb6!

src/ui/component/toast/toast.tsx
0xLodz: We finally removed this feature

0xLodz:
@fredericvilcot @ccamel feedbacks non related to DDD have been taken into account. Do we need to address DDD part  before merging?
Also I am wondering how to allow product team to fully test the feature ? Notifications with CTA cannot currently be triggered from the browser.

src/domain/notification/aggregate.ts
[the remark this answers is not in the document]
fredericvilcot: @0xLodz 
Yes, I'll suggest to leave it like that and immediately do a PR that will normalize these elements.

src/ui/component/toast/toast.tsx
```diff
@@ -1,52 +1,121 @@
...
+  iconName,
+  duration
+}) => {
+  const [isOpen, setIsOpen] = useState(true)
+  const toastRootRef = useRef<HTMLLIElement | null>(null)
+
+  const onClose = useCallback(() => {
+    if (isOpen) {
+      const timeout = setTimeout(() => {
+        clearTimeout(timeout)
```
lolottetheclash, line 65: `clearTimeout(timeout)` should be outside setTimeout, no?

lolottetheclash (changes requested):
Good Job! this feature was not easy with the domain part ;)
Jus some remarks, let me know if you want us to discuss about it !

src/ui/component/notification/notification.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const Notification: FC = () => {
+  const { notifications, dismissNotification } = toEffectfulObject(
+    useNotificationStore(state => ({
+      notifications: state.notifications,
+      dismissNotification: state.dismissNotification
+    }))
+  )
+  const theme = useAppStore(store => store.theme)
+  const notificationData = useMemo(() => notifications(), [notifications])
```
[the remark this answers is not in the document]
ccamel: This one is for me :laughing: 
Well, this tweet explains why: https://twitter.com/nikolaposa/status/1077327810627358721?s=20

Morover, imho, omitting the `get` prefix for getters in entities and APIs (a zustand store is exposed as such) allows the focus to be on the object's properties (which is the domain term) rather than its methods.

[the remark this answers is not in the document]
lolottetheclash: Thanks for these explanation 👍
I think it's confusing to read `notifications()` because it's a mix of const and function (for me), but why not, I'll get used to it 😉

src/ui/page/dataverse/dataspace/governance/governance.tsx
```diff
@@ -76,6 +78,17 @@ export const Governance: FC = () => {
     }
   }, [navigate, navigationPath])
 
+  useEffect(() => {
+    if (!dataspaceLabel) {
```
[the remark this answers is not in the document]
0xLodz: You are right. If this request process well but the resource is not found, we must display the 404 error. We are currently using mocked data retrieved with `getResourceDetails`. I do think if we have a server error processing the request to get dataspace (500), this would be a good use case for warning notification.
We can remove it before merging, then add it once the dataspace is connected to the ontology. 
What do you think @lolottetheclash?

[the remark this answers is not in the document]
lolottetheclash: Yes, the toasts will be for the 500 errors, which we will manage in a second time. 
According to me, currently we only have to manage the 404 👌

src/ui/component/notification/notification.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const notificationTypeForIcon = useMemo(
```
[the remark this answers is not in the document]
0xLodz: Thanks for the suggestion! I agree that the current naming can be confusing. I will update the variable name to `notificationIconByType` for better readability and understanding.

src/ui/component/toast/toast.tsx
```diff
@@ -1,52 +1,121 @@
...
-      return () => {
-        clearTimeout(autoHideDurationTimer)
-      }
+  onDismiss,
+  message,
+  iconName,
+  duration
+}) => {
+  const [isOpen, setIsOpen] = useState(true)
+  const toastRootRef = useRef<HTMLLIElement | null>(null)
```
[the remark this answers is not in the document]
0xLodz: It comes from the ToastRoot from the Radix library :)

src/ui/component/toast/toast.tsx
0xLodz: This is not a conventional but valid approach!
The `clearTimeout(timeout)` is placed before the `onDismiss({ id })` call, which ensures that the timeout is cleared when the parent component causes this component to be unmounted.

src/ui/component/toast/toast.tsx
```diff
@@ -1,52 +1,121 @@
...
-        <div className="okp4-dataverse-portal-toast-header-container">
-          {icon && <div className="okp4-dataverse-portal-toast-icon">{icon}</div>}
-          <p className={classNames(`okp4-dataverse-portal-toast-title ${variant}`)}>{title}</p>
+    <ToastPrimitive.Provider swipeDirection="down">
+      <ToastPrimitive.Root
+        className={classNames('okp4-dataverse-portal-toast-root', {
+          'has-action': action
+        })}
+        duration={duration}
+        onOpenChange={onClose}
```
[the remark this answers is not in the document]
0xLodz: Thanks! will update it

src/domain/notification/aggregate.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export type DismissNotificationInput = {
+  id: NotificationID
+}
+
+type NotificationAction<T> = {
+  reportNotification: (input: ReportNotificationInput<T>) => IO<void>
+  dismissNotification: (input: DismissNotificationInput) => IO<void>
+  notificationWithActionReported: () => IO<boolean>
```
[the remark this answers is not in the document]
0xLodz: Good point, it's a bit confusing. I can rename the function to `hasNotificationWithActionReported`, which indicates more clearly that it checks if the notifications list contains a notification with an action property.

src/ui/component/notification/notification.tsx
ccamel, line 19: Do we need to utilize a memo here? Considering that the notifications are backed by the zustand store, it seems unnecessary, imho. I suggest removing it and directly calling the function at the location where it's needed.

src/domain/notification/entity.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable max-lines-per-function */
+import type { Eq } from 'fp-ts/lib/Eq'
+import type { Option } from 'fp-ts/lib/Option'
+
+export type NotificationID = string
+export type NotificationType = 'success' | 'info' | 'warning' | 'error'
+export type ActionType = 'refresh'
```
ccamel, line 7: Don't declare the action type at this point since the notification is parameterized by `T`. This is to enable the caller to determine the type of actions they wish to facilitate. The domain must remain agnostic to this.

0xLodz: That makes sense, thank you for pointing that out. I have moved the ActionType definition to the Vue component that manages the actions, keeping the domain agnostic to the specific action types.

src/ui/component/notification/notification.tsx
0xLodz: You're right. Zustand helps to avoid unnecessary re-renders by subscribing to the specific parts of the state that a component is using. In this case memoization with `useMemo` is not necessary because the state is already being efficiently managed by Zustand, and any updates to the notifications will trigger a re-render only when required.

lolottetheclash (approved):
Well done 👍

ccamel (approved):
LGTM!  Good job.
It will require an alignment of the domains and some improvements, but it's ok. :+1:

PR #233, opened 2023-05-10
Refactor/error-pages

AnjaMis:
Refactoring the error pages code - creating a common error folder consisting of common scss and translations files and the error pages:  Not Found (404 page) and Internal Error (rendering, code error, error boundary).

Note: to test the internal error, you need to add following line of code in one of the Portal pages: `throw new Error()`


src/ui/page/home/home.tsx
```diff
@@ -5,16 +5,19 @@ import { Explore } from './explore/explore'
 import './home.scss'
 import './i18n/index'
 
-export const Home: FC = () => (
-  <div className="okp4-dataverse-portal-home-page-main">
-    <div className="okp4-dataverse-portal-home-page-block-container community">
-      <Community />
+export const Home: FC = () => {
+  throw new Error()
```
0xLodz, line 9: Do we keep it for testing purpose?

src/ui/page/errors/notFoundError/notFoundError.tsx
```diff
@@ -0,0 +1,17 @@
+import type { FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import { ErrorPage } from '@/ui/component/error/error'
+import '@/ui/page/errors/i18n/index'
+import '@/ui/page/home/i18n/index'
+
+export const NotFoundError: FC = () => {
+  const { t } = useTranslation(['error', 'home'])
```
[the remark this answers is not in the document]
AnjaMis: Ahh yes, true! Removed now :)

src/ui/page/home/home.tsx
AnjaMis: Ooops, completely forgot to remove this🤦🏼‍♀️

0xLodz (approved):
Well done!

lolottetheclash (changes requested):
It's better, but there are some points that could be improved 😉
Here, we notice that the two error pages are almost empty, and that the whole display and style logic is carried by the `Error` component. 
Don't forget that our components must be as dumb as possible.
For example, the `page` property is used by the `Error` component to fully adapt its behavior, this is not good. 
It is up to the page to provide information that the `Error` component consumes stupidly, without being given too much thought.
For example, you assume that there is a set of buttons to display, but what if we have another error page that has no buttons? 
It will still end up with `<div className="okp4-dataverse-portal-error-page-buttons-container"> `which will be empty.
Instead, you could add an `actionComponent` property (the name needs improvement, but it would be nice if it carried the intent, its purpose) that would be optional, and the parent can then provide a react component that will be consumed by the `Error` component and thus displayed.
This will allow to have less code logic in the `Error` component, and more in the page that calls it; same for the css, a good part should stay in each page I think.

lolottetheclash:
As discussed, best and easiest way chosen is to create just a stylesheet common for both pages, and remove the `error` component.

lolottetheclash (changes requested):
Nice work, almost there!
Just a remark though: on `InternalPage` & `notFoundPage`, wouldn't it be possible to have almost the same css?
Because I see that on one side we have style applied on `okp4-dataverse-portal-error-page-text-container` and on the other it is on `okp4-dataverse-portal-error-page-content-wrapper`. I think both could be applied on same classname like in the wrapper.

src/ui/page/home/home.tsx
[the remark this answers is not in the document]
AnjaMis: Ah yes, true 🤦🏼‍♀️

src/ui/page/error/internalError/internalError.scss
```diff
@@ -0,0 +1,35 @@
+@import '@/ui/style/theme.scss';
+@import '@/ui/style/mixins.scss';
+
+@mixin background-image($theme) {
+  background: center / contain url('@/ui/asset/images/internal-error-#{$theme}.webp');
```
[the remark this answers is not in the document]
AnjaMis: Fixed!

lolottetheclash (approved):
Works for me!

ccamel (approved):
LGTM! 👍

src/ui/page/error/internalError/internalError.tsx
```diff
@@ -0,0 +1,49 @@
...
+            <h1 className="okp4-dataverse-portal-error-page-title">{t('sorry')}</h1>
+            <h1>{t('internalError.title')}</h1>
+          </div>
+          <p className="okp4-dataverse-portal-error-page-message">{t('internalError.message')}</p>
+        </div>
+        <div className="okp4-dataverse-portal-error-page-buttons-container">
+          <Button
+            className="okp4-dataverse-portal-error-page-button"
+            label={t('returnToHome')}
+            onClick={handleNavigateTo('/')}
```
fredericvilcot, line 36: It's better to import the `routes` enum and use it as parameter instead of a string that has not contraints

src/ui/page/error/notFoundError/notFoundError.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <h1 className="okp4-dataverse-portal-error-page-title">{t('sorry')}</h1>
+            <h1>{t('notFoundError.title')}</h1>
+          </div>
+          <p className="okp4-dataverse-portal-error-page-message">{t('notFoundError.message')}</p>
+        </div>
+        <div className="okp4-dataverse-portal-error-page-buttons-container">
+          <Button
+            className="okp4-dataverse-portal-error-page-button"
+            label={t('home:home.blocks.explore.label')}
+            onClick={handleNavigateTo('/dataverse')}
```
fredericvilcot, line 34: Same here, you can use `routes` enum 😌

fredericvilcot (changes requested):
Great work! 😌
Just some minor enhancements that enforce typing :)

src/ui/page/error/internalError/internalError.tsx
AnjaMis: You are right, I missed that!

src/ui/page/error/notFoundError/notFoundError.tsx
AnjaMis: Fixed!

fredericvilcot (approved):
Great, thx :)

Issue #219, opened 2023-05-02
[Portal] Field copy in CORE Metadata and Audit Metadata


ChaimaaZaoui:
# Feature: [Portal] Field copy in CORE Metadata and Audit Metadata

> **As a** an end user,
> **I want** to copy a DID in my dashboard 
> **So that** I can use it elsewhere 

### Scenario: Copy DID

**Given that** I am an end user,
**And** I am using a *desktop* or a *mobile* device,
**When** a `copy `icon button is available next to the DID in the UI
**And** I click on the `copy `icon
**Then** the `copy `icon is animated to validate the copy
**And** the concerned DID is copied in my clipboard
**And** the icon goes back to its initial state after 2 seconds 

_Note that the copy snackbar is not required any more_ 

this is applicable now in
- detail page of any dataverse resource 
- governance paragraph 

### Scenario: Failed Copy DID

**Given that** I am an end user,
**And** I am using a *desktop* or a *mobile* device,
**When** a `copy `icon button is available next to the DID in the UI
**And** I click on the `copy `icon
**Then** the `copy `icon is animated to a cross icon to show that the copy is failed 
**And** the icon goes back to its initial state after 2 seconds 

### Mockups

https://user-images.githubusercontent.com/129745194/235853067-3a44c8c8-fc4b-4ad4-ba66-0857fabbfc3e.mov

![image](https://user-images.githubusercontent.com/129745194/236149438-6e1a32b6-10a0-4882-8156-0f6ad9da2edc.png)

### Environment

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206529480) by [Unito](https://www.unito.io)

ChaimaaZaoui:
**Definition of Ready**

- [ ] 
- [ ] - title, description, and value statement
- [ ] - acceptance criteria (scenarios)
- [ ] - mockups (all devices and themes)
- [ ] - labels added
- [ ] - US presented to DEV team (refinement)
- [ ] - attached to an epic
- [ ] - estimation (planning poker)
- [ ] - external dependencies if applicable

ChaimaaZaoui:
[...]

PR #238, opened 2023-05-16
Feat/copy icon animation

ErikssonJoakim:
This PR changes the `CopyToClipBoard` component by indicating copy state through the icon as per specifications. In order to handle the animation timeout I have created two hooks that complement each other: `useTimeout` and `useDebounce`. 

use `useTimeout` to call a function after a given delay. `useDebounce` will call a function only when the timeout delay is passed.

- #219

<details>
<summary>Copy icon animation</summary>

https://github.com/okp4/dataverse-portal/assets/75730728/785a566e-461a-44ac-8b68-75bab9a80024

</details>


src/ui/component/copyToClipboard/copyToClipboard.scss
```diff
@@ -1,9 +1,48 @@
+@mixin hide-icons($icons) {
+  .okp4-dataverse-portal-copy-paste-icon {
```
ccamel, line 2: It's correct, but I'd love to have a check if `$icons` contains values from a specific list to prevent typos.

src/ui/component/copyToClipboard/copyToClipboard.tsx
```diff
@@ -1,32 +1,54 @@
...
 
 export const CopyToClipboard = ({ onCopied, textToCopy }: CopyToClipBoardProps): JSX.Element => {
   const theme = useAppStore(state => state.theme)
+  const [copyState, setCopyState] = useState<CopyState>('copy')
+
+  const resetCopyState = useCallback((): void => {
+    setCopyState('copy')
+  }, [])
+
+  useDebounce(resetCopyState, copyState !== 'copy' ? 2000 : null, [copyState])
```
ccamel, line 23: As an improvement, make the constant `2000` a parameter of the component with a default value.

ccamel (changes requested):

src/ui/component/copyToClipboard/copyToClipboard.scss
ErikssonJoakim: That's a great suggestion! I've implemented it together with a error warning  👍

src/ui/component/copyToClipboard/copyToClipboard.tsx
```diff
@@ -1,32 +1,59 @@
...
+  animationDelay = 2000,
+  onCopied,
+  textToCopy
+}: CopyToClipBoardProps): JSX.Element => {
   const theme = useAppStore(state => state.theme)
+  const [copyState, setCopyState] = useState<CopyState>('copy')
+
+  const resetCopyState = useCallback((): void => {
+    setCopyState('copy')
+  }, [])
```
fredericvilcot, line 26: This is not enough for me 😌

My comments on previous PR about having a hook that handle the copy logic is not implemented here.

All the `copyToClipboard` logic should be abstracted in a hook that will handle the state and the handler function. The component should not have to worry about that, to be dumb and generic as much as possible ;)

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,29 @@
+import { useCallback, useEffect, useRef } from 'react'
+
+type ClearTimeout = () => void
+type ResetTimeout = () => void
+type TimeoutMethods = [ClearTimeout, ResetTimeout]
```
fredericvilcot, line 5:
```suggestion
type UseTimeout = [ClearTimeout, ResetTimeout]
```

More conventional and explicit ;)

fredericvilcot (changes requested):
Great work, you only miss a little subtlety in the abstraction of the copy logic, IMHO.

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useTimeout = (callback: () => void, delay: number | null): TimeoutMethods => {
+  const timeoutRef = useRef<number>()
+
+  const set = useCallback(() => {
+    if (delay) timeoutRef.current = setTimeout(() => callback(), delay)
```
lolottetheclash, line 11: As for clear function, you could also use `&&`

src/ui/component/copyToClipboard/copyToClipboard.tsx
```diff
@@ -1,32 +1,59 @@
...
+
+  useDebounce(resetCopyState, copyState !== 'copy' ? animationDelay : null, [copyState])
```
lolottetheclash, line 28: When component renders, it executes useDebounce even if we haven't click on an icon, I think.

src/ui/component/copyToClipboard/copyToClipboard.tsx
ErikssonJoakim: `<CopyToClipboard />` now checks the `copyState` from the `useCopyToClipboard` hook to know which icon to show, and onClick I use the `handleCopy` that returns a promise telling if the copy succeeded or not.

src/ui/component/copyToClipboard/copyToClipboard.tsx
ErikssonJoakim: I've removed useDebounce as the hook isn't really meant for this use case. I haven't found a way to prevent the set function to not be called when using the hook though. Searching online the hooks I found all have the hook set and cleared on first render.

0xLodz (changes requested):

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useTimeout = (callback: () => void, delay: number | null): UseTimeout => {
+  const timeoutRef = useRef<number | null>(null)
+
+  const set = useCallback(() => {
+    timeoutRef.current = delay && setTimeout(() => callback(), delay)
+  }, [callback, delay])
+
+  const clear = useCallback(() => {
+    timeoutRef.current && clearTimeout(timeoutRef.current)
```
0xLodz, line 15: it would be a good practice to also set timeoutRef.current to null after clearing the timeout. This is to ensure the reference does not hold onto the timeout id once it has been cleared.

ErikssonJoakim: I'm not sure it's needed in this case seeing as it's part of a hook.

0xLodz: It's not necessary, but it's a good practice :)

0xLodz: It's not necessary, but it's a good practice :)

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,29 @@
...
+    clear()
+    set()
+  }, [clear, set])
+
+  useEffect(() => {
+    set()
+    return clear
+  }, [delay, set, clear])
+
+  return [clear, reset]
```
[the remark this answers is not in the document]
ErikssonJoakim: You're right it make more sense for this hook seeing as we might not want to deal with all return values

src/ui/hook/useCopyToClipboard.ts
```diff
@@ -0,0 +1,36 @@
+import { useCallback, useState } from 'react'
+import { useTimeout } from '@/ui/hook/useTimeout'
+
+type HandleCopy = (textToCopy: string) => Promise<void>
+type CopyState = 'idle' | 'success' | 'error'
+
+export const useCopyToClipboard = (
+  copyResetDelay?: number,
+  handler?: (isCopied: boolean) => void
```
fredericvilcot, line 9:
```suggestion
  callback?: (isCopied: boolean) => void
```

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,23 @@
...
+  const clear = useCallback(() => {
+    timeoutRef.current && clearTimeout(timeoutRef.current)
+  }, [])
+
+  useEffect(() => {
+    set()
+    return clear
+  }, [delay, set, clear])
+
+  return { clear }
```
fredericvilcot, line 22:
```suggestion
  return disposerFn
```

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,23 @@
+import { useCallback, useEffect, useRef } from 'react'
+
+type ClearTimeout = () => void
+type UseTimeout = { clear: ClearTimeout }
```
fredericvilcot, line 4:
```suggestion
type UseTimeout = () => void
```

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const useTimeout = (callback: () => void, delay: number | null): UseTimeout => {
+  const timeoutRef = useRef<number | null>(null)
+
+  const set = useCallback(() => {
+    timeoutRef.current = delay !== null ? setTimeout(() => callback(), delay) : null
+  }, [callback, delay])
+
+  const clear = useCallback(() => {
```
fredericvilcot, line 13:
```suggestion
  const disposerFn = useCallback(() => {
```

fredericvilcot (changes requested):
Great, I think you can improve naming to be more explicit, but this is a good job :)

src/ui/hook/useCopyToClipboard.ts
```diff
@@ -0,0 +1,36 @@
+import { useCallback, useState } from 'react'
+import { useTimeout } from '@/ui/hook/useTimeout'
+
+type HandleCopy = (textToCopy: string) => Promise<void>
+type CopyState = 'idle' | 'success' | 'error'
```
ccamel, line 5: Please go further in the definition of the algebraic type, by making the copied text for the `success` state, and the error, for the `error` state.

src/ui/hook/useCopyToClipboard.ts
ccamel, line 9: If you have the copy state available, you can use it directly instead of relying on a callback. Otherwise, there is a mix of the reactive state style and the callback style, imho.

src/ui/hook/useCopyToClipboard.ts
```diff
@@ -0,0 +1,36 @@
...
+    async (textToCopy: string): Promise<void> =>
+      navigator.clipboard
+        .writeText(textToCopy)
+        .then(() => {
+          setCopyState('success')
+          handler?.(true)
+        })
+        .catch((error: Error) => {
+          setCopyState('error')
+          console.error(error.message)
```
ccamel, line 27: Report this in the state.

src/ui/hook/useTimeout.ts
```diff
@@ -0,0 +1,23 @@
+import { useCallback, useEffect, useRef } from 'react'
+
+type ClearTimeout = () => void
+type UseTimeout = { clear: ClearTimeout }
+
+export const useTimeout = (callback: () => void, delay: number | null): UseTimeout => {
```
ccamel, line 6: I don't have time to study the finesse of the implementation but I feel that it is possible to simplify it into just 2 effects, one for the callback, another one for the timeout itself, returning the disposition directly.

ccamel: Ok. As seen with you @ErikssonJoakim, it doesn't seem obvious that the function can be improved. So I propose to keep the current approach which seems to me good. 😌

lolottetheclash (approved):
Not so easy, well done !

src/ui/hook/useCopyToClipboard.ts
```diff
@@ -0,0 +1,52 @@
...
+type CopyStatus = 'idle' | 'success' | 'error'
+
+type Status<T extends CopyStatus> = { status: T }
+type Message = { message: string }
+
+type IdleState = Status<'idle'>
+type SuccessState = Status<'success'> & Message
+type ErrorState = Status<'error'> & Message
+
+type CopyState = IdleState | SuccessState | ErrorState
```
ccamel, line 13: Why not:

```suggestion
export type CopyState = 
  { status: 'idle' } 
  | { status: 'success', message: string } 
  | { status: 'error', message: string };
```

ccamel (changes requested):

src/ui/hook/useCopyToClipboard.ts
```diff
@@ -0,0 +1,52 @@
...
+        .then(() => {
+          setCopyState({
+            status: 'success',
+            message: textToCopy
+          })
+        })
+        .catch((error: Error) => {
+          setCopyState({
+            status: 'error',
+            message: error.message
```
ccamel, line 42: imho, it'd be better to return the error, rather than taking on the responsibility of returning the message. This way, the caller has the opportunity to interpret the error appropriately, without degrading the original information.

ccamel (approved):
Great! Thank you for the job. Much appreciated.

fredericvilcot (approved):
Great job @ErikssonJoakim 🫶

PR #244, opened 2023-05-17
Chore/update dev server script in order to listen on LAN addresses

0xLodz:
In order to make tests on local devices (mobile, tablet, etc...), the dev server needs to listen on LAN addresses.

[See more](https://vitejs.dev/config/server-options.html) about the Vite server configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "pnpm": "^8.3.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


ccamel (approved):
Well spotted. 👍

lolottetheclash (approved):
As discussed before: nice !

Issue #232, opened 2023-05-10
Error page displayed as an intermediate page


ChaimaaZaoui:
Test details : 
Date / Time : 10/05 - 11h36
Browser : Firefox
OS : Windows

Bug description
When I click on `see details` in `.../dataverse` then an error page is displayed a fraction of a second before showing the requested page.

here is a screen recording of the bug, the error page is shown in the second 26
_note that the video is slowed down to x0.25_

ChaimaaZaoui:
[screen-recorder-wed-may-10-2023-11-36-18_dsgsUl2S.webm](https://github.com/okp4/dataverse-portal/assets/129745194/35aab806-4e89-4c82-80f6-b3d9fb4ce727)

fredericvilcot:
@ChaimaaZaoui 
Could we move this issue in TO DO and handle it?

ChaimaaZaoui:
@fredericvilcot it's in to do now :)

PR #250, opened 2023-05-23
Fix/error page flashing on dataverse item card selection

0xLodz:
This PR fixes https://github.com/okp4/dataverse-portal/issues/232.

By fixing this bug, I realised we can get any dataverse item resource, not matter the path we are using.
e. g. `/dataverse/dataset/2` or `/dataverse/service/2` will give the dataverse item of id 2, which shouldn't be the case for the `service` path.
A [PR](https://github.com/okp4/dataverse-portal/pull/255) has been created to address this issue.

### DIFF
--- a/src/ui/page/dataverse/dataset/dataset.tsx
+++ b/src/ui/page/dataverse/dataset/dataset.tsx
@@ -100,7 +100,6 @@ const Dataset: FC = () => {
   const [isLoading, setIsLoading] = useState(true)
 
   useEffect(() => {
-    setIsLoading(true)
     const resourceDetails = id ? getResourceDetails(id) : O.none
     setDataset(
       O.isSome(resourceDetails) && isDataSet(resourceDetails.value) ? resourceDetails : O.none
--- a/src/ui/page/dataverse/dataset/dataset.tsx
+++ b/src/ui/page/dataverse/dataset/dataset.tsx
@@ -1,7 +1,7 @@
+import type { FC } from 'react'
 import { useEffect, useState } from 'react'
 import { useParams } from 'react-router-dom'
-import type { Option } from 'fp-ts/Option'
-import { match, none } from 'fp-ts/Option'
+import * as O from 'fp-ts/Option'
 import { NotFoundError } from '@/ui/page/error/notFoundError/notFoundError'
 import { getResourceDetails } from '@/ui/page/dataverse/dataverse'
 import type { DataverseItemDetails } from '@/ui/page/dataverse/dataverse'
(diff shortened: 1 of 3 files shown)


src/ui/page/dataverse/dataset/dataset.tsx
```diff
@@ -91,15 +92,29 @@ const datasetMetadata: ItemGeneralMetadata[] = [
...
+  const [dataset, setDataset] = useState<O.Option<DataverseItemDetails>>(O.none)
+  const [isLoading, setIsLoading] = useState(true)
 
   useEffect(() => {
-    id && setDataset(getResourceDetails(id))
+    pipe(
+      id,
+      O.fromNullable,
+      O.map(getResourceDetails),
+      O.map(resourceDetails => {
```
ccamel, line 105: `map` is traditionally used for transformations rather than side effects, so I would not use it in this case. I believe that to perform a side effect, it would be more suitable to use an `IO`, and in this specific context where you have to deal with a value or not, an [IOOption](https://gcanti.github.io/fp-ts/modules/IOOption.ts.html).

As it is idiomatic, it would be nice to define this implementation as a reference implementation.

0xLodz: @ccamel thanks for pointing this out. In order to define it as a reference, I would like to have your confirmation about IO in the React context.
`getResourceDetails` currently returns and Option, and probably a task once the vue is wired to the ontology.

Considering `getResourceDetails` is an Option, here is my proposal:
```tsx
const resourceDetails = useMemo(
    () => pipe(id, O.fromNullable, IOO.fromOption, IOO.map(getResourceDetails)),
    [id]
)

useEffect(() => {
  pipe(
    () => setIsLoading(true),
     resourceDetails,
     O.map(setDataspace),
     O.map(() => setIsLoading(false))
  )
}, [resourceDetails])
```

Do we consider state updates within useEffect as side effect as well?
If yes, which reflects probably better the FP principles, we probably want an implementation closer to this one:
```tsx
const resourceDetails = useMemo(
    () => pipe(id, O.fromNullable, IOO.fromOption, IOO.map(getResourceDetails)),
    [id]
)

useEffect(() => {
  pipe(
    IO.of(setIsLoading(true)),
    IO.map(resourceDetails),
    IO.flatMap(
      O.fold(
        () => IO.of(setIsLoading(false)),
        r => {
          setDataspace(r)
          return IO.of(setIsLoading(false))
        }
      )
    )
  )()
}, [resourceDetails])
```

What do you think? Thanks

0xLodz: As we discussed today, we limit the usage of fp-ts to the domain, store, or utility functions. The bug fix has been updated accordingly.

src/ui/page/dataverse/dataset/dataset.tsx
```diff
@@ -91,15 +91,29 @@ const datasetMetadata: ItemGeneralMetadata[] = [
...
+
+const Dataset: FC = () => {
   const { id } = useParams<string>()
-  const [dataset, setDataset] = useState<Option<DataverseItemDetails>>(none)
+  const [dataset, setDataset] = useState<O.Option<DataverseItemDetails>>(O.none)
+  const [isLoading, setIsLoading] = useState(true)
 
   useEffect(() => {
-    id && setDataset(getResourceDetails(id))
+    setIsLoading(true)
```
lolottetheclash, line 103: I think you don't need to set again the loading to true because you already made it into useState initialization ;)

ccamel (approved):

src/ui/page/dataverse/dataset/dataset.tsx
0xLodz: You are right, React reconciles with a different type. Good catch, thanks!

lolottetheclash (approved):
We usually start with a loading state initialized to false, and set it to true when we start loading data. 
But by doing this we still get the flash effect...
It would have been nice to start with loading at false, but let's start that way :)

PR #251, opened 2023-05-24
build(deps): bump fp-ts version to 2.15.0

ccamel:
Self explanatory.

PS:
- this should have been proposed by `dependabot`, but I don't know why it is not
- the [fp-ts official documentation](https://gcanti.github.io/fp-ts/) refers to a version `v2.16`, but it seems not currently available. To be followed.


lolottetheclash (approved):
Thanks for this PR :)

0xLodz (approved):
Thanks!

ErikssonJoakim (approved):

PR #254, opened 2023-05-25
fix: update headScripts async to false in html-config

lolottetheclash:
This PR aims to solve the bug that leads us to a blank & empty page when loading the application.
Indeed, we use a variable in the .env file to connect to the wallet, but which is not necessarily available when we need it.
That's why we pass the loading of the .env file in a synchronous way.


fredericvilcot (approved):

ccamel (approved):
Good catch.

PR #256, opened 2023-05-25
Refactor/harmonize chevrons

0xLodz:
Since we are using two kinds of chevrons in multiple part of the application, we need to keep them in their simple form. This PR prepares the creation of a collapsible component that will use them. The collapsible component is a reusable component needed for filters, articles, menus, and several other elements of the design system.


src/ui/component/toolbar/toolbar.tsx
```diff
@@ -133,11 +131,11 @@ const WalletButton = ({
             `okp4-dataverse-portal-wallet-expand ${isWalletMenuOpen ? 'rotate-down' : 'rotate-up'}`
           )}
         >
-          <Icon name={`expand-down-${theme}`} />
+          <Icon name={`chevron`} />
```
lolottetheclash, line 134: You forgot the `{}` !

lolottetheclash (changes requested):
Nice refactor, thanks !
A tiny oversight to fix and I'm good to go!

src/ui/component/toolbar/toolbar.tsx
0xLodz: Fixed! Thanks

lolottetheclash (approved):
👍

fredericvilcot (approved):

PR #257, opened 2023-05-26
Feat/collapsible component

0xLodz:
## Collapsible Component Creation
This Pull Request introduces the Collapsible component, a flexible and reusable component designed to show or hide content based on user interaction. The user can interact with the trigger button through click, enter, or space key inputs to toggle the visibility of the component's content.

### Purpose
The Collapsible component is primarily developed as a part of the checkbox filter functionality in the dataverse. Its design allows it to be used in various parts of the application such as Governance articles and desktop menus, wallet button interactions, filters, and more. The goal is to create a consistent, reusable interactive element that provides collapsible sections.

### Usage
Both the content of the trigger button and the chevron icon can be customised.
The children represent the content that will be shown or hidden when the trigger button is interacted with.

Here's a simple example of how to use it:
```tsx
<Collapsible
  chevronIconName="chevron"
  triggerClassName="okp4-dataverse-portal-dataverse-filter-toggle-button"
  triggerElement={<h2>My title</h2>}
>
  <div>
    <p>lore ipsum dolor sit amet</p>
    <p>lore ipsum dolor sit amet</p>
  </div>
</Collapsible>
```


src/ui/component/collapsible/collapsible.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const Collapsible: FC<CollapsibleProps> = ({
+  open = false,
+  children,
+  triggerElement,
+  triggerClassName,
+  chevronIconName
+}) => {
+  const [isOpen, setIsOpen] = useState<boolean>(open)
```
ccamel, line 25:
```suggestion
  const [isOpen, setIsOpen] = useState(open)
```

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const handleOpenChange = useCallback((isOpen: boolean) => {
+    setIsOpen(isOpen)
+  }, [])
```
ccamel, line 29: why not:

```suggestion
  const handleOpenChange = setIsOpen
```

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -0,0 +1,52 @@
...
+import classNames from 'classnames'
+import * as RCollapsible from '@radix-ui/react-collapsible'
+import type { IconName } from '@/ui/component/icon/icon'
+import { Icon } from '@/ui/component/icon/icon'
+import type { AssertSubset } from '@/util/type'
+import './collapsible.scss'
+
+type CollapsibleProps = {
+  children: JSX.Element
+  triggerElement: JSX.Element
```
ccamel, line 12: The naming is not consistent. For properties of type `JSX.Elment`, you need to decide whether the property name should be suffixed or not by `Element`.

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useState, useCallback } from 'react'
+import classNames from 'classnames'
+import * as RCollapsible from '@radix-ui/react-collapsible'
+import type { IconName } from '@/ui/component/icon/icon'
+import { Icon } from '@/ui/component/icon/icon'
+import type { AssertSubset } from '@/util/type'
+import './collapsible.scss'
+
+type CollapsibleProps = {
+  children: JSX.Element
```
ccamel, line 11: I prefer the term `content` to `children`, which is what Radix uses. It seems more meaningful to me, imho.

fredericvilcot (approved):

src/ui/component/collapsible/collapsible.tsx
0xLodz: Good suggestion! In this component we don't have any noticeable performance implications of not memoizing. It makes IMO the component easier to read. 

We can even directly use it in the JSX:
```tsx
<RCollapsible.Root onOpenChange={setIsOpen} open={isOpen}>
```

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -0,0 +1,52 @@
...
+  triggerElement: JSX.Element
+  open?: boolean
+  triggerClassName?: string
+  chevronIconName?: AssertSubset<IconName, 'chevron' | 'chevron-sharp'>
```
lolottetheclash, line 15: I think we could let the user choose his own icon to represent the opening/closing of this component.
What's more, if we ever change our mind about which icon to use in our app, it forces us to modify the places where we use this component + modify the props of the component itself. I'm in favor of leaving the choice of icon free, with a default name to simplify its use.

src/ui/component/collapsible/collapsible.tsx
0xLodz: Good point! We can remove the `Element` suffix since the type `JSX.Element` already indicates it's an element. I suggest renaming `triggerElement` to `trigger` for consistency and simplicity.

lolottetheclash (commented):
Thanks for this work, just a remark regarding the icon 😉

src/ui/component/collapsible/collapsible.tsx
0xLodz: `children` is a conventional term in React, but if `content` provides more clarity we can update the props.

src/ui/component/collapsible/collapsible.tsx
0xLodz: Thanks for your suggestion! It brings in fact more flexibility and scalability to the component. In this case, if the user doesn't want any icon, he must explicitly pass a value in order to avoid the fallback on the default parameter, `chevron`.

Is it OK to use `null` as a way to not display any icon, like this:
```suggestion
  iconName?: IconName | null
```

lolottetheclash: I would leave the chevron icon as default, and the user can choose a different one if he prefers.
So I won't add `null` here...

lolottetheclash (approved):
👍

ccamel (approved):
LGTM! Thanks :+1:

PR #258, opened 2023-05-26
Feat/searchbar component

0xLodz:
## SearchBar component creation

This PR introduces the `SearchBar` component, a reusable, controlled component designed to provide an interactive and customisable search input field. The component is built with the intention of being easily integrated into various parts of the application that require search functionality.

### Purpose
The `Searchbar` component is primarily developed as a part of the checkbox filter functionality in the Dataverse. As a controlled component, `SearchBar` provides a predictable data flow and allows for better management of the input state. Thus, other elements can easily rely on search terms controlled by the Searchbar's parent.

### Usage
The`onSearch` function prop that gets triggered whenever the input value changes.

In the following example, `searchTerm` is the current search term that is controlled by the parent component, and `handleSearch` is a function that updates the state of the parent component when the search term changes.

```tsx
<SearchBar
  placeholder="Search..."
  value={searchTerm}
  onSearch={handleSearch}
/>
```

### Visual example
image


ccamel (approved):

lolottetheclash (approved):
👍

lolottetheclash (approved):
👍

fredericvilcot (approved):

PR #260, opened 2023-05-27
Feat/checkbox component

0xLodz:
## Checkbox component creation

This PR introduces the `Checkbox` component, a reusable, controlled component designed to provide an interactive and customisable checkbox input field. The component is built with the intention of being easily integrated into various parts of the application that require selection and filters functionalities.

### Purpose
The `Checkbox` component is primarily developed as a part of the checkbox filter functionality in the Dataverse. As a controlled component, `Checkbox` provides a predictable data flow and allows for better management of the input state. Thus, other elements can easily rely on checbkox state controlled by the Checkbox's parent.

### Usage
The`onCheckedChange` function prop gets triggered whenever the input value changes.

In the following example, `checked` is a value that is controlled by the parent component, and `onCheckedChange` is a function that updates the state of the parent component when the checkbox `checked` attribute change.

```tsx
<Checkbox
  checked={checked}
  value="rhizome"
  name="data-space"
  onCheckedChange={handleCheckedChange}
/>
```

### Visuals states (unchecked, checked, disabled)
image
image

### DIFF
--- a/src/ui/component/checkbox/checkbox.tsx
+++ b/src/ui/component/checkbox/checkbox.tsx
@@ -1,0 +1,54 @@
+import type { FC } from 'react'
+import { useCallback } from 'react'
+import classNames from 'classnames'
+import * as RCheckbox from '@radix-ui/react-checkbox'
+import { Icon } from '@/ui/component/icon/icon'
+import './checkbox.scss'
+
+export type CheckboxOption = {
+  checked: boolean
+  value: string
+  id?: string
+  name?: string
+  disabled?: boolean
+}
+
+type CheckboxProps = CheckboxOption & {
+  onCheckedChange: (checkboxAttributes: Pick<CheckboxOption, 'checked' | 'value' | 'name'>) => void
+}
+
+export const Checkbox: FC<CheckboxProps> = ({
+  checked,
+  value,
+  id = value,
+  name,
+  disabled = false,
+  onCheckedChange
+}) => {
+  const handleCheckedChange = useCallback(() => {
+    if (disabled) return
+
+    onCheckedChange({
+      checked: !checked,
+      value,
+      name
+    })
+  }, [checked, onCheckedChange, disabled, value, name])
+
+  return (
+    <RCheckbox.Root
+      checked={checked}
+      className={classNames('okp4-dataverse-portal-checkbox-main', {
+        checked,
+        disabled
+      })}
+      disabled={disabled}
+      id={id}
+      name={name}
+      onCheckedChange={handleCheckedChange}
+      value={value}
+    >
+      <Icon name="checkbox" />
+    </RCheckbox.Root>
+  )
+}
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@keplr-wallet/types": "^0.11.59",
     "@okp4/eslint-config": "^1.2.0",
     "@radix-ui/react-collapsible": "^1.0.2",
+    "@radix-ui/react-checkbox": "^1.0.3",
     "@radix-ui/react-switch": "^1.0.1",
     "@radix-ui/react-toast": "^1.1.3",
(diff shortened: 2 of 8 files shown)


fredericvilcot (approved):

src/ui/component/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,49 @@
...
+import './checkbox.scss'
+
+export type CheckboxOption = {
+  id: string
+  name: string
+  checked: boolean
+  disabled: boolean
+}
+
+type CheckboxProps = {
```
lolottetheclash, line 15: Here I think you can extend from `CheckboxOption`

src/ui/component/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,49 @@
...
+  id: string
+  name: string
+  checked: boolean
+  disabled: boolean
```
lolottetheclash, line 19: No sure that we need to have disabled mandatory, could you have it optional with a default to false?

src/ui/component/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const Checkbox: FC<CheckboxProps> = ({ id, name, checked, disabled, onCheckedChange }) => {
+  const handleCheckedChange = useCallback(() => {
+    if (disabled) return
+
+    onCheckedChange({
+      checked: !checked,
+      id,
+      name,
+      disabled
```
lolottetheclash, line 31: Not sur you need to return disabled here, because you wrote:
`if (disabled) return ` juste above, so It's automatically `false` here

src/ui/component/checkbox/checkbox.tsx
```diff
@@ -0,0 +1,49 @@
...
+export type CheckboxOption = {
+  id: string
+  name: string
+  checked: boolean
+  disabled: boolean
+}
+
+type CheckboxProps = {
+  id: string
+  name: string
```
lolottetheclash, line 17: is the name important here, as we already have an `id`?

lolottetheclash (changes requested):
Nice, a few comments left below 😉

lolottetheclash (changes requested):
Nice, a few comments left below 😉

src/ui/component/checkbox/checkbox.tsx
0xLodz: Yes, checkbox `name` purpose is to provide a context about the checkbox value or id. [More details here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox).

src/ui/component/checkbox/checkbox.tsx
0xLodz: Yes you are right, I kept it in order to pass the entire Option, making it easier to deal with. I can remove it, but we need to pass `Pick<CheckboxOption, 'checked' | 'id' | 'name'>`, which is probably fine. What do you think ?

src/ui/component/checkbox/checkbox.tsx
0xLodz: Sure, thanks! I realised that I didn't set the `disable` prop to the checkbox. I will fix that.

src/ui/component/checkbox/checkbox.tsx
lolottetheclash: I think that if we don't need a property, we should not pass it :)

src/ui/component/checkbox/checkbox.tsx
lolottetheclash: If it's a context thing, I would have marked name as optional

ccamel (approved):
LGTM! Thx

lolottetheclash (approved):
Great, thanks !

Issue #196, opened 2023-04-24
[Dataverse] Additional filters - Default & multiselect search


ChaimaaZaoui:
# Feature: [Dataverse] Additional filters
> **As an** end-user of the portal on the `.../dataverse `page,
> **I want** to filter the dataverse 
> **So that** only items having the metadata I need are shown

### Scenario 1 : Default display (Desktop, mobile, tablet)
**Given that** I am an end-user navigating the portal
**And** I am using a _desktop_ device,
**When** I am on `.../dataverse `page,
**Then** filters should be displayed in the following order

* Data spaces
* Topics
* Service categories
* Data geographical coverage
* Data temporal coverage
* Data format
* Data license
**And** by default all the filters are open
**And** no item is selected in any filter

### Scenario 2: Filter type : Multiselect search
**Given that** I am an end-user navigating the portal
**When** I am on `.../dataverse `page,
**Then** filters are shown on the left side of the page on desktop **Or** after clicking see filters on mobile
**And** the fields available are :

* an autocomplete `search bar` at the top of the list, 
* the three first items of the list are shown
* the `scroll bar` of the list is on the right
* the list is ordered following an alphabetical order
* By default no selected item

_the concerned filters are :  Data spaces ; Topics; Service categories ;  Data geographical coverage ; Data format ; Data license_

_i : The Autocomplete option provides suggestions while you type into the field and filter the results automatically_

### Scenario 3 : Internationalisation 
ENG: 

* Data spaces
* Topics
* Service categories
* Data geographical coverage
* Data temporal coverage
* Data format
* Data license
* Search a data space
* Search topics
* Search categories
* Search country, regions ...
* Search format
* Search license
* No results found

FR

* Espaces de données
* Sujets
* Catégories de services
* Couverture géographique des données
* Couverture temporelle des données
* Format des données
* Licence des données
* Aucun résultat trouvé
* Rechercher dans un espace de données
* Rechercher des sujets
* Rechercher des catégories
* Rechercher un pays, des régions ...
* Rechercher un format
* Rechercher une licence

DE

* Datenräume
* Themen
* Dienstkategorien
* Geografische Abdeckung der Daten
* Zeitliche Abdeckung der Daten
* Datenformat
* Lizenz der Daten
* Keine Ergebnisse gefunden

* Suche in einem Datenraum
* Themen suchen
* Kategorien suchen
* Suche nach Land, Registern ...
* Suche nach Format
* Suche nach Lizenz

### Mockups
[...]

PR #262, opened 2023-05-29
Feat/dataverse selection filter

0xLodz:
This PR implements https://github.com/okp4/dataverse-portal/issues/196.

### Implementation details
Before wiring the filters to the ontology catalog, we assume that data passed to `SelectionFilter` components are a list of string. When we get more details about it we would be able to update the component signature accordingly.

#### Autocomplete
image

#### folded filters
image

#### Desktop, light theme
image

#### Checkbox label on 2 lines height w/ ellipsis
image

#### No results fallback
image


fredericvilcot (changes requested):
Great work, I gave you some comments about the component itself, that maybe could be less coupled to the checkbox mechanism

src/ui/view/dataverse/component/filters/checkboxFilter/checkboxFilter.tsx
```diff
@@ -0,0 +1,152 @@
...
+type FilterNameProps = {
+  children: string
+}
+export const FilterName: FC<FilterNameProps> = ({ children }) => (
+  <h3 className="okp4-dataverse-portal-dataverse-filter-name">{children}</h3>
+)
+
+type CheckboxFilterProps = {
+  filterName: string
+  filterValues: string[]
```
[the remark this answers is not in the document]
0xLodz: This is related to the `value` attribute of a checkbox element, which for example corresponds to "data analysis", "data management", etc. For these values, `filterName` would be "services categories".

I understand that using "values" is confusing. In order to address this issue, we can probably define a `CheckboxValue` type:

```ts
export type CheckboxValue = string

export type CheckboxOption = {
  checked: boolean
  value: CheckboxValue
  id?: string
  name?: string
  disabled?: boolean
}
```

Then use it for filters:
```tsx
type CheckboxFilterProps = {
  filterName: string
  filterValues: CheckboxValue[]
```

What do you think? Thanks

src/ui/view/dataverse/component/filters/checkboxFilter/checkboxFilter.tsx
```diff
@@ -0,0 +1,152 @@
...
+  })
+
+  const MAX_SEARCH_RESULTS = 10
+
+  const [searchTerm, setSearchTerm] = useState<string>('')
+
+  const [checkboxOptions, setCheckboxOptions] = useState<CheckboxOption[]>(
+    filterValues
+      .map(value => createDefaultCheckboxOption({ name: filterName, value }))
+      .sort((a, b) => a.value.localeCompare(b.value))
```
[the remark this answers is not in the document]
0xLodz: Are you sure that is necessary here? `map` already creates a new array.

src/ui/view/dataverse/component/filters/checkboxFilter/checkboxFilter.tsx
```diff
@@ -0,0 +1,152 @@
...
+)
+
+type CheckboxFilterProps = {
+  filterName: string
+  filterValues: string[]
+  searchPlaceholder: string
+}
+
+// eslint-disable-next-line max-lines-per-function
+export const CheckboxFilter: FC<CheckboxFilterProps> = ({
```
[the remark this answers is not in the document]
0xLodz: Thanks for your feedback. The component has been refactored and renamed with a more generic name: `SelectionFilter`

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+    value,
+    selected: false,
+    disabled: false
+  })
+
+  const [searchTerm, setSearchTerm] = useState<string>('')
+
+  const [filterOptions, setFilterOptions] = useState<SelectionOption[]>(
+    filterValues
+      .map(value => createDefaultFilterOption({ name: filterName, value }))
```
lolottetheclash, line 67: Do you really need the `name` in your `filterOption`? as the name is `filterName`, it is available if you need it. Also, I think you don't use your `option` name in the file.

src/ui/view/dataverse/component/dynamicCheckbox/dynamicCheckbox.tsx
```diff
@@ -0,0 +1,67 @@
...
+import type { SelectionOption } from '@/ui/view/dataverse/component/filters/selectionFilter/selectionFilter'
+import './dynamicCheckbox.scss'
+
+type DynamicCheckboxProps = {
+  name: string
+  value: string
+  checked: boolean
+  highlightedTerm: string
+  disabled?: boolean
+  onCheckedChange: (checkboxOption: Pick<SelectionOption, 'selected' | 'value' | 'name'>) => void
```
lolottetheclash, line 15: I think you can update with an `Omit` ;)

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+type FilterNameProps = {
+  children: string
+}
+export const FilterName: FC<FilterNameProps> = ({ children }) => (
+  <h3 className="okp4-dataverse-portal-dataverse-filter-name">{children}</h3>
+)
+
+type SelectionFilterProps = {
+  filterName: string
+  filterValues: string[]
```
lolottetheclash, line 39: I can see that you sometimes uses `options` to describe `filterValues`, you could have named `filterName`: `label` and `filterValues`: `options`. Not really important, you do as you prefer!

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const handleSearch = useCallback(
+    (searchTerm: string) => {
+      setSearchTerm(searchTerm)
+    },
+    [setSearchTerm]
+  )
+
+  const handleSelectedChange = useCallback(
+    (changedOption: Pick<SelectionOption, 'selected' | 'value' | 'name'>) => {
```
lolottetheclash, line 79: same here: I think you can update with an `Omit` ;)

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+      setFilterOptions(prevFilterOptions => {
+        const optionIndex = prevFilterOptions.findIndex(
+          option => option.value === changedOption.value
+        )
+        if (optionIndex === -1) {
+          return prevFilterOptions
+        }
+        return [
```
lolottetheclash, line 87: I think this works too:

```
      return prevFilterOptions.map(option =>
          option.value === changedOption.value
            ? { ...option, selected: changedOption.selected }
            : option
        )
```

lolottetheclash (changes requested):
Nice work, I wrote you down some feedbacks 😉

src/ui/view/dataverse/component/dynamicCheckbox/dynamicCheckbox.tsx
0xLodz: The goal of using `Pick` is to explicitly show which props from the checkbox are necessary

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+      <span>{t('noResultsFound')}</span>
+    </div>
+  )
+}
+
+type FilterNameProps = {
+  children: string
+}
+export const FilterName: FC<FilterNameProps> = ({ children }) => (
+  <h3 className="okp4-dataverse-portal-dataverse-filter-name">{children}</h3>
```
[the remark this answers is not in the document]
0xLodz: I agree, it makes the component cleaner and provide a most appropriate property!

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+                          name={filterName}
+                          onCheckedChange={handleSelectedChange}
+                          value={value}
+                        />
+                      )
+                  }
+                })
+              ) : (
+                <div className="okp4-dataverse-portal-selection-filter-no-results-wrapper">
+                  <NoResultsFound />
```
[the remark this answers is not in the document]
0xLodz: As you rightly pointed out, the `NoResultsFound` component will be reused in the application. Thus we might want to customize it's parent according to the context and probably keep it as is.

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
```diff
@@ -0,0 +1,144 @@
...
+    [filterOptions, searchTerm, maxSearchResults]
+  )
+
+  return (
+    <div className="okp4-dataverse-portal-selection-filter-main">
+      <Collapsible
+        content={
+          <div className="okp4-dataverse-portal-selection-filter">
+            <SearchBar onSearch={handleSearch} placeholder={searchPlaceholder} value={searchTerm} />
+            <div className="okp4-dataverse-portal-selection-filter-list">
```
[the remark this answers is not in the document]
0xLodz: Sounds a more appropriate name!

src/ui/view/dataverse/component/dynamicCheckbox/dynamicCheckbox.tsx
lolottetheclash: Works for me !

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
0xLodz: You're right, we don't need it to filter the options. The name props will be necessary when the filter is linked to the ontology.

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
0xLodz: Yes `label` might be an appropriate naming too. The`filterName` naming was chosen to match the `name` props of the `SelectionOption`.
The Filter options is more like a `SelectionOption`:
```tsx
type SelectionOption = {
    selected: boolean;
    value: string;
    name: string;
    disabled: boolean;
}
```

src/ui/view/dataverse/component/filters/selectionFilter/selectionFilter.tsx
0xLodz: Yes it works too and it's less verbose. We can update it if you consider it easier to read.

fredericvilcot (approved):

lolottetheclash (approved):
Nice work !

src/ui/page/dataverse/dataverse.tsx
```diff
@@ -433,6 +438,36 @@ const dataverseItems: DataverseItemDetails[] = [
   }
 ]
 
+const selectionFilters = {
```
ccamel, line 441: These data should be fetched from the ontology, as it contains the controlled vocabularies and is therefore the unique source of truth. Will they be taken into account soon?

0xLodz: Yes, the wiring to the ontology is carried out in the next iteration, in a dedicated PR.

PR #266, opened 2023-05-30
Feat/text highlighter component

0xLodz:
## Text Highlighter component
This Pull Request introduces the Text Highlighter component, a reusable component designed to highlight a part of text based on a term.

### Purpose
The Text Highlighter component is primarily developed as a part of the checkbox filter functionality in the dataverse. It will be used for the Filters and the Dropdown and multiselect elements from the design system. The goal is to easily customize a repeated part of a text based on searched/filtered terms.

### Usage

Here's a simple example of how to use it:
```tsx
<TextHighlighter termToHighlight={searchTerm} text={id} />
```
Here is a class that can be nested in any context, e. g. the component parent css class:
```scss
.dataverse-filter-context {
  .text-highlight {
    @include noto-sans(600);
  }
}
```

### Visual example for the Dropdown component
image


src/util/splitTextByTerm/splitTextByTerm.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+  it('returns the original text when the search term is not found', () => {
+    const result = splitTextByTerm('Hello world', 'test')
+    expect(result).toEqual(['Hello world'])
+  })
+
+  it('returns the original text when the search term is empty', () => {
+    const result = splitTextByTerm('Hello world', '')
+    expect(result).toEqual(['Hello world'])
+  })
+})
```
lolottetheclash, line 28: Maybe you should add a test where the original text is empty too? It could also happen

src/ui/component/textHighlighter/textHighlighter.tsx
```diff
@@ -0,0 +1,27 @@
+import type { FC } from 'react'
+import { splitTextByTerm } from '@/util/splitTextByTerm/splitTextByTerm'
+
+type TextHighlighterProps = {
+  text: string
+  termToHighlight: string
+}
+export const TextHighlighter: FC<TextHighlighterProps> = ({
+  text,
+  termToHighlight: textToHighlight
```
lolottetheclash, line 10: I'm not sure to understand why it's renamed here, for me `termToHighlight` was fine but I think that if you prefer `textToHighlight` you could use this name dirctly, no?

lolottetheclash (changes requested):
Nice work, specially to implement it with testing!
I have some remarks, tell me if you need explanations ;)

src/ui/component/textHighlighter/textHighlighter.tsx
0xLodz: Good catch thanks! The component prop type was renamed but I forgot to rename the prop accordingly.

src/ui/component/textHighlighter/textHighlighter.tsx
```diff
@@ -0,0 +1,27 @@
...
+}) => {
+  const parts = splitTextByTerm(text, textToHighlight)
+
+  return (
+    <>
+      {parts.map((part, index) =>
+        part.toLowerCase() === textToHighlight.toLowerCase() ? (
```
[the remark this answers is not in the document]
0xLodz: We can provide such option. I do think most of the time the user quickly want to search for a term. He doesn't care, even sometimes doesn't know where are the capitalized letters in the text to search. What do you think?

src/ui/component/textHighlighter/textHighlighter.tsx
```diff
@@ -0,0 +1,27 @@
...
+          <span className="text-highlight" key={index}>
```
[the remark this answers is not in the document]
0xLodz: It's seems better in term of maintainability to not rely on CSS cascade, although it prevents from defining style at upper level, e. g. page or app level.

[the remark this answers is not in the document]
0xLodz: It's seems better in term of maintainability to not rely on CSS cascade, although it prevents from defining style at upper level, e. g. page or app level.

src/util/splitTextByTerm/splitTextByTerm.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { splitTextByTerm } from './splitTextByTerm'
+
+describe('splitTextByTerm', () => {
```
ccamel, line 3: What if:

```typescript
splitTextByTerm('hello world', '\x') // 💣💥
```

src/util/splitTextByTerm/splitTextByTerm.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+const isNonEmpty = (s: string): boolean => !S.isEmpty(s)
+
+export const splitTextByTerm = (text: string, term: string): string[] =>
+  pipe(
+    term,
+    O.fromPredicate(isNonEmpty),
+    O.fold(
+      () => [text],
+      t => pipe(text.split(new RegExp(`(${t})`, 'gi')), A.filter(isNonEmpty))
```
ccamel, line 23: RegExp can raise an exception if the given regular expression is not well formed. If you want to use the RegExp, you must escape all special characters in string.

src/ui/component/textHighlighter/textHighlighter.tsx
ccamel: Ideally, a comparison function should be provided as part of the props, so that you can "scan" the term to highlight in the text. But the implementation is a bit tricky, from what I can see.

src/ui/component/textHighlighter/textHighlighter.tsx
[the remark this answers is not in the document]
0xLodz: According to the product team, lowercase and uppercase have no impact on searches.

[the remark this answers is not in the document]
ccamel: Ideally, a comparison function should be provided as part of the props, so that you can "scan" the term to highlight in the text. But the implementation is a bit tricky, from what I can see.

And the default implementation would be the string comparison while ignoring case (of course), so that the user doesn't have to worry about it, most of the time.

src/util/splitTextByTerm/splitTextByTerm.ts
0xLodz: Good catch, thanks! Special characters are now escaped using the function `escapeRegExp` in [4a6638e](https://github.com/okp4/dataverse-portal/pull/266/commits/4a6638e01011090eb5e56f07b1a59cbfaa0c4a58).

src/util/splitTextByTerm/splitTextByTerm.spec.ts
0xLodz: Thanks for pointing this out! This is not an error that can be caught or handled at runtime, because it prevents your code from running at all.
If the search term comes from some source other than a user input field, it's the responsibility of the code that generates the search term to ensure that it's a valid string.

Invalid escape sequences are now handled from the function `replaceInvalidEscapeSequences` in [4a6638e](https://github.com/okp4/dataverse-portal/pull/266/commits/4a6638e01011090eb5e56f07b1a59cbfaa0c4a58).

ccamel (approved):
LGTM!

lolottetheclash (approved):
Nice! well done 👌

PR #267, opened 2023-06-02
fix: update dataverse items data in explore page

lolottetheclash:
As the dataverse elements are now retrieved from the ontology, the fake dataverse element ids on the explore page are no longer relevant.
When the user clicks on the "View details" button of one of the 4 dataverse item cards in the explore part of the home page, we receive a 404 page because the existing id is not linked to any dataverse item.
This ontology query has not yet been implemented, so the fake dataverse items ids have been updated with those from the ontology.


0xLodz (approved):

fredericvilcot (approved):
👍

Issue #268, opened 2023-06-02
[Dataverse] Apply FIlters


lolottetheclash:
On staging:

If the connection is slow and you quickly press (for example) services then dataspaces, only the services will be displayed (not services and dataspaces).

Expected behaviour :
we want to see each result associated with the selected filters. 
In our example we want to see services and dataspaces

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206526525) by [Unito](https://www.unito.io)

PR #269, opened 2023-06-04
Fix/dataverse data fetching

0xLodz:
## Bug fix for fetch operation on slow or unstable network connections using abortController signal

This pull request addresses #268. It provides a solution to a bug that surfaces when a user is operating with a slow or unstable internet connection. Prior to this fix, when the user switched for example between **Datasets** and **Services** by selecting these filters quickly, he would not receive expected data for services. This issue resuled from a condition in the `loadDataverse` domain method where a fetch operation was not performed if the data was [still loading](https://github.com/okp4/dataverse-portal/blob/5446ad01ea1788cb683f6d82c14892d7455dbcc4/src/domain/dataverse/domain.ts#L141).

The implementation in this PR uses the AbortController signal to ensure that a new fetch operation is performed each time a request is made, regardless of the loading state of the previous fetch operation. This effectively ensures that every user interaction that should trigger a fetch operation will do so and receive the appropriate data, even under conditions of an unstable or slow network connection.

This PR introduces a new utility: `createAbortableFetch`, which creates a new AbortController for each fetch request. It aborts the previous fetch request when a new one is made, ensuring that the fetch operation corresponds to the most recent user interaction.
Additionally, the `handleAbortError` utility is used to handle the error thrown when an abort operation is performed.

EDIT: A command `cancelDataverseLoading` has been created following the @fredericvilcot feedback.

### DIFF
--- a/src/domain/dataverse/command.ts
+++ b/src/domain/dataverse/command.ts
@@ -7,6 +7,8 @@ export type ByTypeFilterInput = DataverseElementType | 'all'
 export type Command = {
   // Load the dataverse elements from outside
   loadDataverse: () => Task<void>
+  // Cancel the loading of the dataverse elements
+  cancelDataverseLoading: () => Task<void>
   // Set the language in which the dataverse should be translated to.
   setLanguage: (newLng: string) => IO<void>
   // Filter the dataverse by one or multiple element types
--- a/src/domain/dataverse/command.ts
+++ b/src/domain/dataverse/command.ts
@@ -7,8 +7,6 @@ export type ByTypeFilterInput = DataverseElementType | 'all'
 export type Command = {
   // Load the dataverse elements from outside
   loadDataverse: () => Task<void>
-  // Cancel the loading of the dataverse elements
-  cancelDataverseLoading: () => Task<void>
   // Set the language in which the dataverse should be translated to.
   setLanguage: (newLng: string) => IO<void>
   // Filter the dataverse by one or multiple element types
--- a/src/domain/dataverse/domain.ts
+++ b/src/domain/dataverse/domain.ts
@@ -10,7 +10,6 @@ import { devtools } from 'zustand/middleware'
 import type { StoreApi } from 'zustand/vanilla'
 import { createStore } from 'zustand/vanilla'
 import type { ForgetType } from '@/util/type'
-import { FetchAbortError } from '@/util/fetch/fetch'
 import type { Dataverse } from './entity'
 import type { DataversePort } from './port'
 import type { ByTypeFilterInput, Command, DataverseElementType } from './command'
@@ -136,7 +135,7 @@ export const storeFactory = (
         loadDataverse: (): T.Task<void> =>
           pipe(
             T.fromIO(() => get().data),
-            T.flatMap(data =>
+            T.chain(data =>
               pipe(
                 T.fromIO(() =>
                   set(state => ({ data: { ...state.data, isLoading: true, error: O.none } }))
@@ -153,16 +152,16 @@ export const storeFactory = (
                   e =>
                     pipe(
                       e,
-                      O.fromPredicate(e => e instanceof FetchAbortError),
+                      O.fromPredicate(e => e.name === 'AbortError'),
                       O.fold(
-                        // do not update state if fetch was aborted
-                        () => T.of(undefined),
                         () =>
                           T.fromIO(() =>
                             set(state => ({
                               data: { ...state.data, isLoading: false, error: O.some(e) }
                             }))
-                          )
+                          ),
+                        // do not update state if fetch was aborted
+                        () => T.of(undefined)
                       )
                     ),
                   r =>
@@ -179,7 +178,14 @@ export const storeFactory = (
                 )
               )
             )
-          )
+          ),
+        cancelDataverseLoading: (): T.Task<void> =>
+          T.fromIO(() => {
+            pipe(
+              O.guard(get().data.isLoading),
+              O.map(() => gateway.abortRequest())
+            )
+          })
       })),
       {
         anonymousActionType: 'Aggregate',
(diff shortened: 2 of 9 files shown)


fredericvilcot (changes requested):
Great work @0xLodz , that is a good first step for the resolution of this issue! 👍

But, if we want to go further and take a little height in terms of architecture, I think certain elements are missing.

In a BDD-oriented hexagonal architecture, the priority is decoupling and in particular between the `infrastructure`, `controller` and `domain` layers.

Here, you systematically abort the request in the gateway itself, which tends to bring business logic into the infrastructure layer.
To respect the principles of hexagonal architecture, this action must be included in the contract between the domain and the outside world, thanks in particular to the `port`.

Thus, the responsibility for this action is transferred to the `domain` itself, ie it is up to it to decide whether or not to abort the request (according to business rules).

If we want to go even further, this action should also be carried by a `command` that can be triggered by any `controller`.

For example, how would you do if you wanted to offer the user a button to cancel the request, and thus cancel it only on user action? (for example if it takes too long and the user wants to cancel it)

This is where the BDD part comes into play: through an undo `command`, you express a user behavior, a feature.

In this example, we can assume that we want to go straightforward by only expressing a `port` method in a first time.

src/domain/dataverse/command.ts
```diff
@@ -7,7 +7,7 @@ export type ByTypeFilterInput = DataverseElementType | 'all'
 export type Command = {
   // Load the dataverse elements from outside
   loadDataverse: () => Task<void>
-  // Set the language in which the dataverse should be translated to.
+  // Cancel the loading of the dataverse elements
```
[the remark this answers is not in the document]
0xLodz: Oops! Sorry about this change. I fixed it back to the initial comment.

src/domain/dataverse/port.ts
```diff
@@ -17,4 +17,5 @@ export type DataversePort = {
     offset: number,
     filters: RetrieveDataverseQueryFilters
   ) => TE.TaskEither<Error, RetrieveDataverseResult>
+  abortRequest: () => void
```
fredericvilcot, line 20: This naming is a bit confusing because you express in the `port` how the dataverse must be retrieved, (by aborting a **Request**)

IMHO, the `port` only reflects the contract between the logic (hexagon) and the outside (infra), so you should not couple it with the inner implementation of the `gateway`.

To illustrate this, will the name `abortRequest` be appropriated if the `gateway` only parses a local file to retrieve the `dataverse`? No, because there is no request in this case. You'll probably want to cancel the readable stream.

So, I'll suggest to only express here the notion of cancelling the retrieval of the `dataverse`, that is the only purpose of the `port`.

src/domain/dataverse/domain.ts
```diff
@@ -137,46 +137,44 @@ export const storeFactory = (
...
               pipe(
-                data,
-                O.fromPredicate(() => !data.isLoading),
-                O.match(
-                  () => T.of(undefined),
-                  data =>
+                T.fromIO(() =>
+                  set(state => ({ data: { ...state.data, isLoading: true, error: O.none } }))
+                ),
+                T.chainFirst(() => T.fromIO(() => gateway.abortRequest())),
```
fredericvilcot, line 143: IMHO, you don't need to call the gateway in another `IO` because you are already in.
```suggestion
                T.chainFirst(() => gateway.abortRequest()),
```

fredericvilcot (changes requested):

src/domain/dataverse/domain.ts
0xLodz: Yes the `gateway.abortRequest()` synchronous operation is an `IO` effect. However I think it still needs `T.fromIO` to be wrapped as a Task in order to be used in the Task context.
Your suggestion was my initial attempt, but Typescript came at me with this error: `Type 'void' is not assignable to type 'Task<unknown>'.ts(2322)`

src/util/fetch/fetch.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+
+describe('fetch utilities', () => {
+  beforeEach(() => {
+    fetchMock.resetMocks()
+  })
+  describe('createAbortableFetch', () => {
+    test('aborts previous request', async () => {
+      fetchMock.mockResponses(
+        [JSON.stringify({}), { status: 200 }],
+        [JSON.stringify({}), { status: 200 }]
```
lolottetheclash, line 12: I'm not sure I understand correctly this part: is it supposed to mock each request response? 
if it's the case, I think the first one is cancelled so it can't have the 200 status 🤔

fredericvilcot (approved):
LGTM!

src/util/fetch/fetch.spec.ts
0xLodz: Yes it is, and you're right, it would be more appropriate to set a `0` status, or no status. I updated the test.

src/domain/dataverse/domain.ts
fredericvilcot: I see the point, LGTM

lolottetheclash (changes requested):
As discussed with @0xLodz a part of the test is not relevant as we can delete the `fetchMock.mockResponses` and see the test pass anyway.
Also, if we change the second url into `https://examsdffsdfsdfdsfple.com`, the test fails so it's really fetching the url.

0xLodz:
> As discussed with @0xLodz a part of the test is not relevant as we can delete the `fetchMock.mockResponses` and see the test pass anyway. Also, if we change the second url into `https://examsdffsdfsdfdsfple.com`, the test fails so it's really fetching the url.

@lolottetheclash thank you for spotting this error! You are correct, mocks must be enabled at config level to be applied globally, or at test level, which has been done [here](https://github.com/okp4/dataverse-portal/compare/209a0e6b1eac87883d1ec059c56fe4258f4b941f..bb5b7983358adeb05c120dde3d926ffe409e0372).
Mocks and assertions has been updated in order to avoid false positive.

lolottetheclash (approved):

Issue #235, opened 2023-05-16
[Share data] Select a storage service - `A local folder` general display


ChaimaaZaoui:
> **As a** end-user with an authenticated wallet or not,
> **I want** to reference a dataset in the dataverse,
> **So that** my dataset could be used to create knowledge. 

### Scenario 1: "Select a storage service" general display
**Given that** I am an end user,
**And** I am using a _desktop_ device,
**When** I am in the step "Select a storage service"
**Then** the page is displayed as follows

* **Text**  I SHARE A DATASET
* **Figure** Stepper 
* **Text** Select a storage service
* **Text** MY DATASET IS IN
* **Tab** section 
-- A local folder _(Active tab)_
-- An external storage _(Inactive tab)_

### Scenario 2: `A local folder` general display
**Given that** I am an end user,
**And** I am using a _desktop_ device,
**When** I am in the step "Select a storage service"
**And** I selected the `A local folder` tab
**Then** the tab content is displayed as follows

* **Search bar** Search a storage service
* Service Card
* **Button**`next step >`
* A scroll bar with an infinite scroll and a loader

### Scenario 3: `A local folder` Service card display
**Given that** I am an end user,
**And** I am using a _desktop_ device,
**When** I am in the step "Select a storage service"
**And** I selected the `A local folder` tab
**Then** the service card contains 

* **Text** Service
* **Text** Service title
* **Text**_Publisher : Service publisher_
* **Button**`Select`

### Scenario 4: Default display
**Given that** I am an end user,
**And** I am using a _desktop_ device,
**When** I am in the step "Select a storage service"
**And**`A local folder` tab is selected
**Then** by default

* No service is selected
* `Next step >` is disabled

### Scenario 5: Internationalisation
ENG
> I Share a dataset
> Select a storage service
> My dataset is in
> A local folder
> An external storage
> Search a storage service
> Service 
> Publisher
> Select // Selected
> Next step

FR
> Je partage un jeu de données
> Sélectionner un service de stockage
> Mon jeu de données se trouve dans
> Un dossier local
> Un stockage externe
> Rechercher un service de stockage
> Service 
> Éditeur 
> Sélectionner // Sélectionné
> Prochaine étape

DE
> Ich gebe einen Datensatz frei
> Wählen Sie einen Speicherdienst
> Mein Datensatz befindet sich in
> Einem lokalen Ordner
> Ein externer Speicher
> Suche nach einem Speicherdienst
> Dienst 
> Verleger
[...]

PR #278, opened 2023-06-08
Feat: add storage service selection page

lolottetheclash:
This Pr implements the UI part of the share dataset page [1st step](https://github.com/okp4/dataverse-portal/issues/235).

All that remains is to modify the scrollbar color, which will be implemented at a later date to take advantage of Loïc's mixin, which is contained in another PR.
The `Tabs` component has been coded to contain only 2 tabs (it has not been designed to contain more), which is why it has been integrated into the `views` folder (to be shared in the `dataverse`) and not in the `components` folder, which is supposed to contain generic elements.

### DIFF
--- a/src/ui/component/card/card.tsx
+++ b/src/ui/component/card/card.tsx
@@ -13,11 +13,15 @@ export type CardProps = {
   children: React.ReactElement
   background?: CardBackground
   mainClassName?: string
+  onClick?: () => void
 }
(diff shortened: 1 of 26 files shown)


ccamel (approved):
Great! :+1:

0xLodz (approved):
Great! LGTM!

PR #300, opened 2023-06-14
Fix: update sparql query

lolottetheclash:
This PR aims to search dataverse item title with fallback if the desired language is not provided.
Also, dataverse data provided to display dataverse item details has been updated to fit ontology.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206526232) by [Unito](https://www.unito.io)

### DIFF
--- a/src/infra/dataverse/sparql/sparqlGateway.ts
+++ b/src/infra/dataverse/sparql/sparqlGateway.ts
@@ -39,7 +39,7 @@ export const sparqlGateway: DataversePort = {
       PREFIX datasetMetadata: <https://ontology.okp4.space/metadata/dataset/>
       PREFIX dataspaceMetadata: <https://ontology.okp4.space/metadata/dataspace/>
       PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
-      SELECT ?id ?metadata ?title ?type ?publisher ?topic (COALESCE(?filteredPrefLabel, ?fallbackPrefLabel) as ?prefLabel)
+      SELECT ?id ?metadata ?title ?type ?publisher ?topic ?prefLabel
       WHERE {
         {
           ?id rdf:type core:Service .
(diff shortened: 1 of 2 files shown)


fredericvilcot (approved):
Great 😊

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -58,16 +58,23 @@ export const sparqlGateway: DataversePort = {
...
         ?metadata core:describes ?id .
-        ?metadata core:hasTitle ?title .
-        FILTER langMatches(lang(?title),'${language}')
+        OPTIONAL { 
+          ?metadata core:hasTitle ?title .
+          FILTER langMatches(lang(?title),'${language}')
+        }
+        OPTIONAL { 
+          ?metadata core:hasTitle ?title .
+        }
```
ccamel, line 67: Seems redundant. Is it required?

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -58,16 +58,23 @@ export const sparqlGateway: DataversePort = {
...
         ?metadata core:hasPublisher ?publisher .
         OPTIONAL { 
-          ?topic skos:prefLabel ?filteredPrefLabel .
-          FILTER langMatches(lang(?filteredPrefLabel),'${language}')
+          ?topic skos:prefLabel ?prefLabel .
+          FILTER langMatches(lang(?prefLabel),'${language}')
+        }
+        OPTIONAL { 
+          ?topic skos:prefLabel ?prefLabel .
```
ccamel, line 74: Same here.

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -58,16 +58,23 @@ export const sparqlGateway: DataversePort = {
...
         }
-        ?topic skos:prefLabel ?fallbackPrefLabel .
       }
-      ORDER BY ?title
+      ORDER BY asc(str(?title))
```
ccamel, line 77: why is it required?

antho31 (approved):
OK now I begin to understand how to query with SPARQL \o/

src/infra/dataverse/sparql/sparqlGateway.ts
lolottetheclash: Otherwise, the query will first group by the language desired by the user, then sort. 
As a practical consequence, the results concerning the fallback language come afterwards and are not sorted.
This is the only way to have the sorting performed on the title value, whatever the language.

src/infra/dataverse/sparql/sparqlGateway.ts
lolottetheclash: If the language filter returns nothing, this ensures a fallback value, if there is no fallback value and the value is not found, then it returns no result.

fredericvilcot:
@ccamel 
Are you OK with these changes?

src/infra/dataverse/sparql/sparqlGateway.ts
ccamel: :thinking: Did you try it? IMHO, this current query might not perform as expected. It first focuses on a particular language, but then it considers all languages on the second otional. This means it could return titles in any language, not just the one we're interested in.

I guess you could have three separate optional sections. One for the language you're interested in, one for english (`en`), and one for no language. And you can smartly design your SPARQL template to skip the english section when you're already targeting english.

fredericvilcot:
@lolottetheclash 
I think we could close this PR as these SparQL improvements have never been approved and we need more help to make them relevant. What do you think?

PR #302, opened 2023-06-16
fix(infra): remove useless contains method

fredericvilcot:
This PR enhances the way the `byType` filter is handled in the `sparql` query and that by replacing the unnecessary `CONTAINS` expression by a simple match on the URI literal value.


lolottetheclash (approved):
🚀

0xLodz (approved):

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -22,11 +22,11 @@ export const sparqlGateway: DataversePort = {
     offset: number,
     filters: RetrieveDataverseQueryFilters
   ): TE.TaskEither<Error, RetrieveDataverseResult> => {
-    const buildContainsExpression = (filter: DataverseElementType): string =>
-      `contains(str(?type), "${filter}" )`
+    const buildStrExpression = (filter: DataverseElementType): string =>
+      `str(?type) = str(core:${filter})`
```
ccamel, line 26: I would remove the operator `str` on both sides to make a plain URI comparison:

```suggestion
    const buildStrExpression = (filter: DataverseElementType): string =>
      `?type = core:${filter}`
```

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -22,11 +22,11 @@ export const sparqlGateway: DataversePort = {
...
 
     const buildQueryFilter = (index: number, acc: string, cur: DataverseElementType): string =>
-      index === 0 ? buildContainsExpression(cur) : `${acc} || ${buildContainsExpression(cur)}`
+      index === 0 ? buildStrExpression(cur) : `${acc} || ${buildStrExpression(cur)}`
 
     const byTypeFilter = (f: DataverseElementType[]): string =>
```
ccamel, line 31: you can simplify by just considering that the filter is a join of eq expressions (provided by `buildStrExpression`):

```suggestion
       const byTypeFilter = (filters: DataverseElementType[]): string =>
        filters.map(buildStrExpression).join(' || ')
```

ccamel (changes requested):
Minor comments - we can look together if you want.

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: Great, thank you for this improvement !

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: Yes, you're right, I thought it was better to compare literals over URIs but I was mistaken.

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: I implemented it, but in `fp-ts` way 😜

ccamel: > I implemented it, but in `fp-ts` way 😜

far better, indeed ☺️

ccamel (approved):
👍

Issue #236, opened 2023-05-16
[Resource referencing] Multi-Step Form with Stepper - V1


ChaimaaZaoui:
As a user, 
I want to have a clear understanding of my progress in a multi-step form by visualizing it through a stepper component, 
and I want to be able to navigate through the form using Next and Previous buttons.

### Scenario 1: Navigating a Multi-Step Form with Stepper and Navigation Buttons

Given that I am accessing a Resource referencing process, 
When I start filling out the fields, 
Then I should see a stepper component that visually represents the steps of the process
and indicates the current step I am on.

### Scenario 2: Moving forward
Given that I am on a specific step of the Resource referencing process, 
When I have completed the required fields on the current step, 
Then the `Next `button should be active which allows me to proceed to the next step.

Given that I am on a specific step of the Resource referencing process, 
When I have not completed the required fields on the current step, 
Then the `Next `button remain inactive and non-clickable.

Given that I am on the last step of the Resource referencing process, 
Then the `Next `button should be hidden.

### Scenario 3: Moving backward
Given that I am on a step other than the first step of the Resource referencing process, 
When I want to review or modify the information I provided on the previous step, 
Then I should see a "Previous" button, which allows me to navigate back to the previous step. 

When I click the `Previous` button, 
Then I should be taken back to the previous step, 
and my previously entered information should be preserved.

Given that I am on the first step of the Resource referencing process, 
When I am reviewing or modifying the information provided on the first step, 
Then the "Previous" button should be hidden, as there is no previous step to navigate back to.

### Scenario 4: Internationalisation

Eng
previous
next"

Fr
précédent
suivant

DE
zurück
weiter

#### Desktop

![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/71dfb917-3c54-4c46-b03f-46170171fb7c)

![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/e2f65159-90bc-432f-9e60-fbc6512f0540)

![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/fd956171-c628-4783-81bb-ef4b3a1cd175)

### Environment

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206527369) by [Unito](https://www.unito.io)

ChaimaaZaoui:
Add the navigation part, with the next and previous button

[...]

PR #312, opened 2023-06-19
Feat/create stepper progress component

0xLodz:
## Stepper Progress component creation

This PR introduces `StepperProgress` and `ProgressBar` components, designed to provide an overview of the progression for [the Stepper](https://github.com/okp4/dataverse-portal/issues/236).

The entire Stepper component is developed in parallel in [this PR](https://github.com/okp4/dataverse-portal/pull/314).
Creating two separates PR make it easier to read and review.

### Behavior

A video from [this comment](https://github.com/okp4/dataverse-portal/issues/236#issuecomment-1582172446) provides more details for the Stepper Progress behavior.

#### When Next step is reached
- the previous step changes from background blue (active) to validate (green).
- the next step changes from gray (default status) to blue (active), via a translation on the X axis.
#### When the previous step is reached
- the step that was active returns to its previous state via a reverse translation on the X axis
- the step that becomes active changes to blue with a transition on the background

### Usage
The`StepperProgress` will be a part of the Stepper component from [this technical specification](https://github.com/okp4/dataverse-portal/issues/236#issuecomment-1578757452).

### Visual example
image

### Testing
A test can be performed from the branch of the PR #314 , that includes both commits.

### DIFF
--- a/src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
+++ b/src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
@@ -1,0 +1,20 @@
+import classNames from 'classnames'
+import type { FC } from 'react'
+import type { StepStatus } from '../stepperProgress'
+import './progressBar.scss'
+
+type ProgressBarState = 'active' | Extract<StepStatus, 'complete' | 'incomplete'>
+export type ProgressBarTransition = 'none' | 'backward' | 'forward'
+
+type ProgressBarProps = {
+  state: ProgressBarState
+  transition: ProgressBarTransition
+}
+
+export const ProgressBar: FC<ProgressBarProps> = ({ state, transition }) => (
+  <div
+    className={classNames('okp4-dataverse-portal-progress-bar-main', state, transition, {
+      'delayed-transition': transition === 'forward'
+    })}
+  />
+)
--- a/src/ui/component/stepper/stepperProgress/stepperProgress.tsx
+++ b/src/ui/component/stepper/stepperProgress/stepperProgress.tsx
@@ -1,0 +1,68 @@
+import type { FC } from 'react'
+import type { ProgressBarTransition } from './progressBar/progressBar'
+import { ProgressBar } from './progressBar/progressBar'
+import './stepperProgress.scss'
+
+type StepId = string
+export type StepStatus = 'error' | 'disabled' | 'complete' | 'incomplete'
+
+type Step = {
+  id: StepId
+  order: number
+  status: StepStatus
+}
+
+type StepperProgressProps = {
+  steps: Step[]
+  activeStepId: StepId
+  previousActiveStepId: StepId
+}
+
+const getProgressBarTransition = ({
+  stepOrder,
+  activeStepOrder,
+  previousActiveStepOrder
+}: {
+  stepOrder: number
+  activeStepOrder: number
+  previousActiveStepOrder: number
+}): ProgressBarTransition => {
+  if (stepOrder === activeStepOrder && previousActiveStepOrder < stepOrder) {
+    return 'forward'
+  }
+  if (stepOrder === previousActiveStepOrder && activeStepOrder < stepOrder) {
+    return 'backward'
+  }
+  return 'none'
+}
+
+const findStep = (steps: Step[], stepId: StepId): Step =>
+  steps.find(({ id }) => id === stepId) ?? steps[0]
+
+export const StepperProgress: FC<StepperProgressProps> = ({
+  steps,
+  activeStepId,
+  previousActiveStepId
+}) => {
+  const activeStep = findStep(steps, activeStepId)
+  const previousActiveStep = findStep(steps, previousActiveStepId)
+
+  return (
+    <div className="okp4-dataverse-portal-stepper-progress-main">
+      {steps.map(({ id, status, order }) => (
+        <ProgressBar
+          key={id}
+          state={id === activeStepId ? 'active' : status === 'complete' ? 'complete' : 'incomplete'}
+          transition={getProgressBarTransition({
+            stepOrder: order,
+            activeStepOrder: activeStep.order,
+            previousActiveStepOrder: previousActiveStep.order
+          })}
+        />
+      ))}
+      <span className="okp4-dataverse-portal-stepper-progress-step-count">
+        {activeStep.order + 1 + '/' + steps.length}
+      </span>
+    </div>
+  )
+}
--- a/src/ui/component/stepper/stepperProgress/progressBar/progressBar.scss
+++ b/src/ui/component/stepper/stepperProgress/progressBar/progressBar.scss
@@ -1,0 +1,73 @@
+@import '@/ui/style/theme.scss';
+@import '@/ui/style/mixins.scss';
+
+$transition-duration: 600ms;
+
+.okp4-dataverse-portal-progress-bar-main {
+  @include with-theme {
+    width: 40px;
+    height: 6px;
+    border-radius: 3px;
+    position: relative;
+    overflow: hidden;
+    background-color: rgba(themed('text'), 0.12);
+    
+    &.delayed-transition {
+      transition-delay: $transition-duration;
+    }
+    
+    &.active {
+      background-color: themed('primary-color-variant-3');
+      transition-duration: $transition-duration;
+      transition-property: background-color;
+    }
+    
+    &.complete {
+      background-color: themed('success');
+      transition-duration: $transition-duration;
+      transition-property: background-color;
+    }
+  
+    &::before {
+      content: '';
+      position: absolute;
+      inset: 0;
+      background-color: themed('primary-color-variant-3');
+      transform: translateX(-100%);
+    }
+
+    &.none::before {
+      transform: translateX(-100%);
+    }
+    
+    &.forward::before {
+      animation: progress-forward $transition-duration forwards;
+    } 
+    
+    &.backward::before {
+      animation: progress-backward $transition-duration forwards;
+    }
+  }
+}
+
+@keyframes progress-backward {
+  from {
+    transform: translateX(0%);
+  }
+  
+  to {
+    transform: translateX(-100%);
+  }
+}
+    
+  
+@keyframes progress-forward {
+  from {
+    transform: translateX(-100%);
+  }
+
+  to {
+    transform: translateX(0%);
+  }
+}
+  
--- a/src/ui/component/stepper/stepperProgress/stepperProgress.scss
+++ b/src/ui/component/stepper/stepperProgress/stepperProgress.scss
@@ -1,0 +1,18 @@
+@import '@/ui/style/fonts.scss';
+@import '@/ui/style/theme.scss';
+@import '@/ui/style/mixins.scss';
+
+.okp4-dataverse-portal-stepper-progress-main {
+  @include with-theme {
+    display: flex;
+    gap: 8px;
+    align-items: center;
+
+    .okp4-dataverse-portal-stepper-progress-step-count {
+      font-size: 16px;
+      @include noto-sans(400);
+      color: rgba(themed('text'), 0.4);
+      line-height: 22px;
+    }
+  }
+}


src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const ProgressBar: React.FC<ProgressBarProps> = ({
+  color = 'grey',
+  transition = 'none'
+}) => {
+  return (
+    <div
+      className={classNames('okp4-dataverse-portal-progress-bar-main', color, transition)}
+      style={{
+        transitionDelay: transition === 'forward' ? '600ms' : '0ms' // 600ms must match the transitions and animations durations in progressBar.scss file
```
lolottetheclash, line 21: Isn't it easier to add `transition-delay: 600ms;` in `forward` class scss?

src/ui/component/stepper/utils.ts
```diff
@@ -0,0 +1,26 @@
+import type {
+  ProgressBarColor,
+  ProgressBarTransition
+} from './stepperProgress/progressBar/progressBar'
+import type { StepStatus } from './stepperProgress/stepperProgress'
+
+export const getProgressBarColor = ({ valid, active }: StepStatus['status']): ProgressBarColor =>
+  active ? 'blue' : valid ? 'green' : 'grey'
+
+export const getProgressBarTransition = ({
```
lolottetheclash, line 10: If I understand well, these methods are only used in `StepperProgress` component, so I would have let them in `StepperProgress.tsx` as they are not shared

src/ui/hook/usePrevious.ts
```diff
@@ -0,0 +1,9 @@
+import { useRef, useEffect } from 'react'
+
+export const usePrevious = <T>(value: T): T | undefined => {
```
lolottetheclash, line 3: I'm wondering if it's really useful to make it as a hook, I think you could keep your ref in the stepper progress file.

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
```diff
@@ -0,0 +1,51 @@
...
+  status: {
+    valid: boolean
+    active: boolean
+  }
+}
+
+type StepperProgressProps = {
+  stepsStatus: StepStatus[]
+  activeStepIndex: number
+  totalSteps: number
```
lolottetheclash, line 20: I think you can tell from the length of the `stepsStatus`, or is it a case where you need it ?

lolottetheclash (changes requested):
Thanks for this work! I got some remarks that we can discuss of course :)

src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
0xLodz: The transition delay is related to the `background-color` transition for the progress bar element, and doesn't affect the pseudo element animation. For more clarity, we might define a dedicated class instead of inlining the property?

src/ui/component/stepper/utils.ts
0xLodz: The purpose was to isolate types that are not directly related to the progress bar, e. g. `StepStatus`. We can definitely move it in into the progressBar file if this doesn't bring that much value.

src/ui/hook/usePrevious.ts
0xLodz: This is a widely used hook pattern that might be reusable somewhere else in the app. Until we need it somewhere else, keeping it in the stepper progress file is fine too.

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
```diff
@@ -0,0 +1,51 @@
...
+}) => {
+  const previousActiveStepIndex = usePrevious(activeStepIndex) ?? 0
+
+  return (
+    <div className="okp4-dataverse-portal-stepper-progress-main">
+      {stepsStatus.map(({ id, status }, index) => (
+        <ProgressBar
+          color={getProgressBarColor({
+            valid: status.valid,
+            active: status.active
```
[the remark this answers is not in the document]
0xLodz: Yes it does. Status of the step can be either **valid** and **active**. If the step is not **active**, the progress bar color fallbacks depending on the **valid** status.

src/ui/component/stepper/utils.ts
0xLodz: The purpose was to isolate types that are not directly related to the progress bar, e. g. `StepStatus`. We can definitely move it in into the progressBar file if this doesn't bring that much value.

src/ui/hook/usePrevious.ts
```diff
@@ -0,0 +1,9 @@
...
+  const ref = useRef<T>()
+  useEffect(() => {
+    ref.current = value
+  })
+  return ref.current
```
fredericvilcot, line 8: IMHO, it's better to update `ref` only when `value` changes, not each time the hook is invoked.
```suggestion
  useEffect(() => {
    ref.current = value
  }, [value])
  return ref.current
```

src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
```diff
@@ -0,0 +1,25 @@
+import classNames from 'classnames'
+import './progressBar.scss'
+
+export type ProgressBarColor = 'grey' | 'blue' | 'green'
```
fredericvilcot, line 4: Instead of coupling the `ProgressBar` behavior with style considerations (`color`), I would have handled the `status` as a prop that is then taken into account in the component's stylesheet.

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ProgressBar } from './progressBar/progressBar'
+import { getProgressBarColor, getProgressBarTransition } from '../utils'
+import { usePrevious } from '@/ui/hook/usePrevious'
+import './stepperProgress.scss'
+
+type StepId = string
+
+export type StepStatus = {
+  id: StepId
+  status: {
```
fredericvilcot, line 11: You can keep it simple here, I guess by using a `status` enum:
```suggestion
  status: "active' | 'valid' | 'error'
```

src/ui/component/stepper/utils.ts
```diff
@@ -0,0 +1,26 @@
+import type {
+  ProgressBarColor,
+  ProgressBarTransition
+} from './stepperProgress/progressBar/progressBar'
+import type { StepStatus } from './stepperProgress/stepperProgress'
+
+export const getProgressBarColor = ({ valid, active }: StepStatus['status']): ProgressBarColor =>
```
fredericvilcot, line 7: Seems useless if you directly use the status as `ProgressBar` prop

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
```diff
@@ -0,0 +1,51 @@
...
+    valid: boolean
+    active: boolean
+  }
+}
+
+type StepperProgressProps = {
+  stepsStatus: StepStatus[]
+  activeStepIndex: number
```
fredericvilcot, line 19: Why using here an index instead of the `stepId`?

fredericvilcot (changes requested):

src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
lolottetheclash: thanks for your explanation, yes I think we should avoid inline style to remain as consistant as possible 👍

src/ui/hook/usePrevious.ts
lolottetheclash: You're right it makes sense 👍

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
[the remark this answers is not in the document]
lolottetheclash: ok, but I expected the status to take only one of several values in a union 🤔

0xLodz:
Following team feedback, I have completely overhauled the PR's content and description. Thanks!

lolottetheclash (approved):
Nice job, thanks !

src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
```diff
@@ -0,0 +1,20 @@
+import classNames from 'classnames'
+import type { FC } from 'react'
+import type { StepStatus } from '../stepperProgress'
+import './progressBar.scss'
+
+type ProgressBarState = 'active' | Extract<StepStatus, 'complete' | 'incomplete'>
```
fredericvilcot, line 6: I can't see clearly the advantage of using here an `Extract` utility type over as example a `Pick` one, in terms of intention.

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
```diff
@@ -0,0 +1,73 @@
+import type { FC } from 'react'
+import type { ProgressBarTransition } from './progressBar/progressBar'
+import { ProgressBar } from './progressBar/progressBar'
+import './stepperProgress.scss'
+
+export type StepId = string
+export type StepStatus = 'error' | 'disabled' | 'complete' | 'incomplete'
+
+export type StepData = {
```
fredericvilcot, line 9: I will have simplified by only `Step`, that is self-explanatory enough

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
```diff
@@ -0,0 +1,73 @@
...
+  if (stepOrder === previousActiveStepOrder && activeStepOrder < stepOrder) {
+    return 'backward'
+  }
+  return 'none'
+}
+
+const findStepData = (stepsData: StepData[], stepId: StepId): StepData => {
+  const stepData = stepsData.find(({ id }) => id === stepId)
+  if (!stepData) {
+    throw new Error(`Step data not found for step id: ${stepId}`)
```
fredericvilcot, line 42: Ok, we can throw an error if this rule is not respected and that will be caught by the `ErrorBoundary`.

On another side, we could prevent the app from crashing and handle a fallback behaviour that will be displayed by the `progressBar`.

Could be discussed 😌

fredericvilcot (commented):
Good job, just some minor interrogations 😌

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
0xLodz: A `Step` type is created for the Step component in the stepper PR, that's why I tried to create an other one for step data. We can change `StepData` to `Step` and probably update the other PR `Step` type to `StepElement` or something similar. What do you think?

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
0xLodz: If we are using this function in the contexte of the stepper, it won't fail. It was mostly done to make stepper components more readable and avoiding to rely on the `as` keyword.

We can arbitrary fallback on the first step for instance. What do you think? Thanks

src/ui/component/stepper/stepperProgress/progressBar/progressBar.tsx
fredericvilcot: Sorry, I didn't notice the type of the step statuses! 
LGTM !

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
fredericvilcot: You can scope the `Step` type to the `stepperProgress` component or create a sub-type inherited from `Step` in `stepper` component, because you handle the same properties here no? (`id`, `order`, `status`)

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
fredericvilcot: Yes, I'll prefer to handle fallback behavior over throwing, thanks 😌

src/ui/component/stepper/stepperProgress/stepperProgress.tsx
0xLodz: Yes! We can definitely create such relationship using a sub-type. It makes the component easier to read. Thanks!

ccamel (approved):

fredericvilcot (approved):
Great thx :)

fredericvilcot (approved):
Great Thx 🤓

Issue #245, opened 2023-05-19
[Share data] Select a storage service - `A local folder` search feature


ChaimaaZaoui:
> **As a** As a user in the step "Select a storage service" in `A Local folder`
> **I want** to be able to perform a dynamic text search for item titles so that 
> **So that** I can quickly find relevant storage service based on my search queries.

### Scenario: Dynamic search

**Given that** I am in `A local folder` tab
**And** I am using a *desktop* device,

**When** I type in a text in the search bar, 
**Then** the service cards should dynamically update based on the entered text.
**And** the search results should include all services whose **titles** contain the entered text
**And** the search results should be updated in real-time as I continue to type or modify the text in the search bar.

**When** no matching item titles are found, 
**Then** the search results should display a message indicating "No results found"

#### Desktop
![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/2aced55c-21a6-469f-86f8-21b7f15c9b0a)

### Environment

- staging URL: <https://app.staging.okp4.network/>

## Additional informations

None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206526926) by [Unito](https://www.unito.io)

ChaimaaZaoui:
DOR
- [x] title, description, and value statement
- [x] acceptance criteria (scenarios)
- [x] mockups (all devices and themes)
- [x] labels added
- [x] US presented to DEV team (refinement)
- [ ] attached to an epic
- [ ] estimation (planning poker)
- [x] external dependencies if applicable

ChaimaaZaoui:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/DEV-board--63e210cc9b1f2f73ee3f4e5f/issues/okp4/dataverse-portal/245?planning-poker) with Zenhub @0xLodz @amimart @AnjaMis @ccamel @ErikssonJoakim @fredericvilcot @lolottetheclash

ChaimaaZaoui:
Add : matching in order and not any character

ChaimaaZaoui:
@fredericvilcot Please share the technical solution

ccamel:
AFAIS, we should use the SPARQL [`contains` predicate](https://www.w3.org/TR/sparql11-query/#func-contains) for filtering. However, to ensure functional and performance aspects, it is necessary to conduct tests on our [fuseki](https://jena.apache.org/documentation/fuseki2/) instance.

[...]

PR #313, opened 2023-06-19
Feat/share dataset service research

lolottetheclash:
This PR implements this [ticket](https://github.com/okp4/dataverse-portal/issues/245) which allows the user to search for a service by entering text in the search bar.


src/domain/dataverse/domain.ts
```diff
@@ -90,15 +96,29 @@ export const storeFactory = (
...
+              pipe(
+                x,
+                O.fromNullable,
+                O.match(
+                  () => IO.of(''),
+                  filter => IO.of(filter.value)
+                )
+              )
+            )
+          ),
```
ccamel, line 121: A little too complicated for my liking. Let's try to simplify using the `flow` operator instead of `pipe`, and by avoiding the unecessary `flatMap`, where a simple `map` is sufficient:

```suggestion
        byPropertyFilter: () =>
          pipe(
            () => get().data.filters.byProperty,
            IO.map(
              flow(
                O.fromNullable,
                O.match(
                  () => '',
                  filter => filter.value
                )
              )
            )
          ),
```

src/domain/dataverse/domain.ts
```diff
@@ -110,7 +130,7 @@ export const storeFactory = (
               }
             }
           })),
-        setLanguage: (newLng: string): IO<void> =>
+        setLanguage: (newLng: string): IOType<void> =>
```
ccamel, line 133:
```suggestion
        setLanguage: (newLng: string): IO.IO<void> =>
```

(if you want to set the return type explicitly)

src/domain/dataverse/domain.ts
```diff
@@ -179,7 +199,25 @@ export const storeFactory = (
...
+              filters: {
+                ...state.data.filters,
+                byProperty: pipe(
+                  newFilter.value,
+                  O.fromPredicate(v => !!v.length),
+                  O.match(
+                    () => null,
+                    () => newFilter
+                  )
+                )
```
ccamel, line 217: Option shouldn't be used for conditional branching, use [B.match](https://gcanti.github.io/fp-ts/modules/boolean.ts.html#match) instead; the semantic is clearer and more expressive, imho.

I think the following should do the job:

```suggestion
                byProperty: pipe(
                  !!newFilter.value.length,
                  B.match(
                    () => null,
                    () => newFilter
                  )
                )
```

ccamel (changes requested):
A few minor comments :wink:

src/domain/dataverse/domain.ts
```diff
@@ -4,15 +4,21 @@ import * as A from 'fp-ts/Array'
 import * as O from 'fp-ts/Option'
 import * as T from 'fp-ts/Task'
 import * as TE from 'fp-ts/TaskEither'
-import type { IO } from 'fp-ts/IO'
+import type { IO as IOType } from 'fp-ts/IO'
```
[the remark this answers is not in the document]
lolottetheclash: Yes you're right

src/domain/dataverse/domain.ts
lolottetheclash: much better, thanks !!!

src/domain/dataverse/domain.ts
```diff
@@ -179,7 +198,24 @@ export const storeFactory = (
...
+            data: {
+              ...state.data,
+              dataverse: [],
+              filters: {
+                ...state.data.filters,
+                byProperty: pipe(
+                  !!newFilter.value.length,
+                  B.match(
+                    () => null,
+                    () => newFilter
```
0xLodz, line 213: You can also make it more readable by avoiding the double bang operator:
```suggestion
                  newFilter.value,
                  S.isEmpty,
                  B.match(
                    () => newFilter,
                    () => null
```

src/ui/page/share/dataset/steps/serviceStorageSelection/serviceStorageSelection.tsx
```diff
@@ -32,20 +35,38 @@ export const ServiceStorageSelection: FC = () => {
...
-    console.log(searchTearm)
-  }, [])
+  const handleServiceSearch = useCallback(
+    (searchTearm: string) => {
+      setSelectedService(null)
+      setByPropertyFilter({ property: 'title', value: searchTearm })()
+      loadDataverse()()
+    },
+    [loadDataverse, setByPropertyFilter]
+  )
```
0xLodz, line 69: Even though we are aborting results of previous calls, it might be better to debounce the service search function in order to reduce the API calls frequency.

lolottetheclash: I'd actually thought about it and tested it, the problem being that, since the component is controlled, the value to be displayed in the search bar is returned by the parent. 
Its value is `byPropertyFilter()()`, which is calculated by `handleServiceSearch` (which calls `setByPropertyFilter`).
As a result, when debouncing, the letters typed by the user take a long time to appear in the search bar.

src/domain/dataverse/domain.ts
lolottetheclash: Yes we can also make it this way, for me it was shorter and easy to understand with the `!!`but I can make the change :)

fredericvilcot (approved):
Great!

src/ui/page/share/dataset/steps/serviceStorageSelection/serviceStorageSelection.tsx
0xLodz: Yes, for a controlled component, you don't have the choice of passing the parent's value. This value can come from a local state of the parent, which doesn't need to be synchronised with `byPropertyFilter()()`. I think it's completely fine for our case.

We can address this later on.

0xLodz (approved):
🚀

PR #314, opened 2023-06-19
Feat/stepper

0xLodz:
## Stepper component creation

This PR implements the [Stepper component](https://github.com/okp4/dataverse-portal/issues/236), an interactive and reusable tool for step-by-step navigation. This component enables transitions between steps while also displaying the progress and status of each step.
This component is designed for seamless integration across various application segments requiring step functionality, such as data sharing processes.

## Step API
The Stepper component takes the `steps` props, which is a list of Step objects with a `content` props, that contains customisable JSX :
```tsx
type Step = {
  id: StepId
  order: number
  status: StepStatus
  content: JSX.Element
}
```
⚠️ Steps validation is postponed to a second iteration. Thus it prevents to block the development and integration of the steps.

## Testing
1. create a `stepperExemple.tsx` file in the stepper folder `src/ui/component/stepper/stepperExemple.tsx`, with the following content:

```tsx
/* eslint-disable max-lines-per-function */
/* eslint-disable react/jsx-no-bind */

import { useMemo, useState } from 'react'
import type { StepElement } from './stepper'
import { Stepper } from './stepper'

export const StepperExample: FC = () => {
  const [input1Value, setInput1Value] = useState('')
  const [input2Value, setInput2Value] = useState('')

  const steps: StepElement[] = useMemo(
    () => [
      {
        id: '1',
        status: 'incomplete',
        content: (
          <div>
            <h2 style={{ color: 'white' }}>Step One</h2>
            <label>
              <input
                onChange={(e): void => setInput1Value(e.target.value)}
                type="text"
                value={input1Value}
              />
            </label>
          </div>
        )
      },
      {
        id: '2',
        status: 'incomplete',
        content: (
          <div>
            <h2 style={{ color: 'white' }}>Step Two</h2>
            <label>
              <input
                onChange={(e): void => setInput2Value(e.target.value)}
                type="text"
                value={input2Value}
              />
            </label>
          </div>
        )
      },
      {
        id: '3',
        status: 'incomplete',
        content: (
          <div>
            <h2 style={{ color: 'white' }}>Step Three</h2>
            <p style={{ color: 'white' }}>Input 1: {input1Value}</p>
            <p style={{ color: 'white' }}>
              Input 2: <span>{input2Value}</span>
            </p>
            <p style={{ color: 'white' }}>
              Lorem ipsum dolor sit amet consectetur adipisicing elit. Sint reiciendis temporibus
              necessitatibus saepe quibusdam quaerat? Nemo doloribus vero repellendus, et dolore
              fuga veritatis modi debitis dolor minus, temporibus atque mollitia?
            </p>
          </div>
        )
      }
    ],
    [input1Value, input2Value]
  )

  return <Stepper steps={steps} />
}


```

2. Add the `<StepperExample />` component anywhere in the app.


src/ui/component/stepper/useStepper.ts
```diff
@@ -0,0 +1,97 @@
...
+  nextStep: () => void
+  previousStep: () => void
+  activeStepId: StepId
+  previousActiveStepId?: StepId
+}
+
+type ActiveStepsId = {
+  current: StepId
+  previous?: StepId
+}
```
fredericvilcot, line 17: I can't really see why this is needed in the hook because you can always compute the previous active step based upon the active step, there is no need to memoize the `previous` active step through a state.

Moreover, `ActiveStepsId` suggests that there are several steps in memory.

src/ui/component/stepper/stepper.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <div className="okp4-dataverse-portal-stepper-content">{activeStep.content}</div>
+        </CSSTransition>
+      </TransitionGroup>
+
+      <StepperActions
+        activeStepId={activeStepId}
+        nextStep={nextStep}
+        previousStep={previousStep}
+        steps={steps}
+      />
```
fredericvilcot, line 59: Does `StepperAction` really need the whole `steps` as prop?

fredericvilcot (approved):
Thanks for your work @0xLodz .

I will approve this PR because we need to implement urgently this feature and we have no time to make changes on it.

But, I have to say there is some structurant elements to rework in the hook and the `Stepper` component at architecture level (contract interfaces).

Let's improve it in a very next future, thanks 🙏

src/ui/component/stepper/useStepper.ts
0xLodz: We can definitely improve the state naming. Regarding the previous active step, it's kept in memory because we don't have a way to figure which one was the previous one from the steps.

src/ui/component/stepper/stepper.tsx
0xLodz: Good point! We can only pass the `isFirstStep` and `isLastStep`.

src/ui/component/stepper/useStepper.ts
```diff
@@ -0,0 +1,97 @@
+import { useCallback, useState } from 'react'
+import type { Step, StepElement, StepId, StepStatus } from './stepper'
```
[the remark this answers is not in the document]
0xLodz: Thank you for the detailed explanation. I will update it accordingly

lolottetheclash (approved):

Issue #242, opened 2023-05-16
[Share data] Select your data - Select a file from the explorer 


ChaimaaZaoui:
# Feature: [Share data] Select your data - Select a file from the explorer 

As a user, I want to be able to browse my file explorer to select files/folder and upload them into the service.

### Scenario: Select a file from device file explorer 

**Given that** I am an end user,
**When** I am in _Select your data_ step, 
**And** I click on the `browse` button
**Then** a menu opens with 2 options : `Select folders` and `Select files`) 
**When I click on one of the 2 options, 
**Then** the file explorer in my device is launched and I can select either files or folders according to my choice. 

**When** I select one or multiple files/folders
**And** validate the selection from the file/folders explorer
**Then** The files/folders should be automatically selected into the system

**When** the files/folders are successfully uploaded
**Then** the files/folders are to be displayed in a list next to the drag and drop section in a list with scroll bar.

**When** new files/folders are added 
**Then** they are shown on the top of the file list

Error management is in : #246

### Scenario: Internationalisation

EN
Browse
Select files
Select folder

FR
Parcourir
Sélectionner des fichiers
Sélectionner un dossier

GE
durchsuchen
Dateien auswählen
Ordner auswählen

#### Desktop

![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/39e8a602-3d13-41cd-a2e5-dddc29966399)

https://github.com/okp4/product-kanban/assets/115076387/b9322403-fc4d-4684-8263-ba063de3a01f

### Environment

- staging URL: <https://app.staging.okp4.network/>

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206527089) by [Unito](https://www.unito.io)

0xLodz:
## Techincal specs
Please refer the [technical specs](https://github.com/okp4/dataverse-portal/issues/240#issuecomment-1580491377) from the issue #240.

ErikssonJoakim:
**Technical Specification**

| Use Case | API | description |
| - | - | - |
| User clicks on Browse | [input type="file"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file) | On click the browser opens. With multiple attribute assigned the input field accepts multiple files  |
[...]

Issue #241, opened 2023-05-16
[Share Data] Select your data - File Drag and drop


ChaimaaZaoui:
As a user, I want to be able to drag and drop files into the service for easy uploading.

### Scenario 1: File drag and drop

**Given that** I am an end user,
**When** I am in _Select your data_ step, 
**Then** a designated area where I can drop files is available,
**And** I should be able to select one or multiple files from my device by clicking and dragging them into the designated area.

**When** dragging the files over the area
**Then** the area takes a hover state

**When** dropping the files
**Then** The files should be automatically selected 

**When** the files are successfully selected
**Then** the files are to be displayed in a list next to the drag and drop section in a list with scroll bar.

**When** new files are added 
**Then** they are shown on the top of the file list

Error management is in : #246 

#### Desktop
![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/ea93a7ae-3490-45e6-ae70-2b52243bbf9a)

![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/710013ad-4930-40d9-9c77-b848e9102662)

### Environment

- staging URL: <https://app.staging.okp4.network/>

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206527159) by [Unito](https://www.unito.io)

ErikssonJoakim:
[drag and drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API)
[drop event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/drop_event)

| Use Case                                                   | Attribute                                         | Description |
| ----------                                                  | -----------                            | ----------- |
| Drag files/folders over area hover state | onDragEnter, onDragLeave | We can use the API to know when to add and remove classNames in className state |
| Drop files/folders & selecting them        | onDrop                                  | onDrop event contains the files dropped as datatransfer objects with a getAsFile method. The selection can be handled in the onDropHandler with a useState hook |

The list will use the selected state to populate itself.

ChaimaaZaoui:
Techspec domain to be done

MdechampG:
[...]

Issue #246, opened 2023-05-19
[Share Data] Select your data - Selection Error


ChaimaaZaoui:
As a user, I want to be able to drag and drop files/folders into the service for easy uploading and know if an error is encountered.

### Scenario 1 : Error while file selection
**Given that** I am an end user,
**And** I am in _Select your data_ step, 
**When** I have dropped files/folders
**And** an error is encountered while the selection
**Then** an error snack-bar is displayed with the following message:
>Selection Failed
>An error occurred while selecting the file. Please try again later.

###Scenario Internationalisation 
ENG
>Selection Failed
>An error occurred while selecting the file. Please try again later.

FR
>Error lors de la sélection
>Une erreur s'est produite lors de la sélection du fichier. Veuillez réessayer plus tard.

DE
>Auswahl fehlgeschlagen
>Bei der Auswahl der Datei ist ein Fehler aufgetreten. Bitte versuchen Sie es später noch einmal.

#### Desktop

![image](https://github.com/okp4/dataverse-portal/assets/129745194/536775e2-b34f-48b7-a80a-7bcda3874136)

ChaimaaZaoui:
- [ ]  title, description, and value statement
- [ ] acceptance criteria (scenarios)
- [ ] mock-ups (all devices and themes)
- [ ] labels added
- [ ] US presented to DEV team (refinement)
- [ ] attached to an epic
- [ ] estimation (planning poker)
- [ ] external dependencies if applicable

MdechampG:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/DEV-board--63e210cc9b1f2f73ee3f4e5f/issues/okp4/dataverse-portal/246?planning-poker) with Zenhub @0xLodz @amimart @ccamel @ErikssonJoakim @fredericvilcot @lolottetheclash

fredericvilcot:
@ChaimaaZaoui 
Below is a screen record of the requested behavior (because you can't directly test this feature): 
https://images.zenhubusercontent.com/593979806/e7ff55b9-eb35-4f8a-a998-fe8a044d688c/enregistrement_de_l__cran_2023_07_05___16_53_21.mov

ChaimaaZaoui:
Screen recording by fred => Success

PR #316, opened 2023-06-20
Feat/file drag and drop component

ErikssonJoakim:
This PR introduces the <FilePicker /> component that will be used in the share data page.
- #242 
- #241 

**Demo video**
https://github.com/okp4/dataverse-portal/assets/75730728/25a87419-2450-4eca-ae72-d0afa2107f22

Use the following code and replace `<ServiceStorageSelection />` with `<DataSelection />` to quickly test the component:

**dataSelection.tsx**
```tsx
import '@/ui/page/share/i18n/index'
import './dataSelection.scss'
import { FilePicker } from '@/ui/component/filePicker/filePicker'

export const DataSelection: FC = (): JSX.Element => {
  const { t } = useTranslation(['common', 'share', 'publisher'])

  return (
    <div className="okp4-dataverse-portal-share-dataset-page-data-selection-container">
      <h2>{t('share:share.dataset.selectStorage')}</h2>
      <h3>{t('share:share.dataset.location')}</h3>
      <div className="okp4-dataverse-portal-share-dataset-page-data-selection-left-wrapper">
        <FilePicker />
      </div>
      <div className="okp4-dataverse-portal-share-dataset-page-data-selection-right-wrapper">
        <div draggable style={{ height: '100px', backgroundColor: 'green' }}></div>
      </div>
    </div>
  )
}

```

**dataSelection.scss**
```scss
@import '@/ui/style/fonts.scss';
@import '@/ui/style/theme.scss';
@import '@/ui/style/mixins.scss';

.okp4-dataverse-portal-share-dataset-page-data-selection-container {
  @include with-theme {
    @include twelve-column-layout;
    grid-template-rows: auto auto 1fr;

    h2 {
      @include grid-item($colStart: 1, $colEnd: -1, $rowStart: auto, $rowEnd: auto);
      @include noto-sans(700);
      font-size: 28px;
      line-height: 28px;
      letter-spacing: 0.4px;
      color: themed('text');
    }

    h3 {
      @include grid-item($colStart: 1, $colEnd: -1, $rowStart: auto, $rowEnd: auto);
      @include noto-sans(700);
      font-size: 16px;
      line-height: 24px;
      letter-spacing: 0.4px;
      color: themed('text');
      margin-bottom: 16px;
    }

    .okp4-dataverse-portal-share-dataset-page-data-selection-left-wrapper {
      @include grid-item(1, 6, auto, auto);
    }
    .okp4-dataverse-portal-share-dataset-page-data-selection-right-wrapper {
      @include grid-item(6, -1, auto, auto);
    }
  }
}

```


src/ui/component/filePicker/filePicker.tsx
fredericvilcot, line 1: Globally, I would say that it's working well but this component seems not generic at all, despite the fact it is part of the global `component` folder.

I mean you didn't parametrized anything (multiple, error, etc..) and I am not a big fan to directly call the domain commands in the component.

IMHO, you could just add some props to handle some custom parameters and the callback functions, that will be easy to implement and will add the first step towards the future design system.

By the way, how errors are handled in the component?

src/ui/component/dragAndDrop/dragAndDrop.tsx
```diff
@@ -0,0 +1,58 @@
...
+    onChange?.(dragDropState)
+  }, [onChange, dragDropState])
+
+  return (
+    <div
+      onDragEnter={handleDragEnter}
+      onDragLeave={handleDragLeave}
+      onDragOver={preventDefaultAndStopPropagation}
+      onDrop={handleDrop}
+      style={{ width: '100%', height: '100%' }}
```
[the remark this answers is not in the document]
ErikssonJoakim: I did contemplate doing this but IMO the benefits of a dedicated stylesheet are questionable for this component. I don't see it evolve to carry any other style and this way we improve (although slightly) the readability of the component.

There might be considerations that I haven't taken into account but if you agree I prefer to leave it as is.

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -18,15 +20,24 @@ export const Collapsible: FC<CollapsibleProps> = ({
   open = false,
   content,
   trigger,
+  contentClassName,
+  rootClassName,
   triggerClassName,
   iconName = 'chevron'
 }) => {
+  const [triggerInitWidth, setTriggerInitWidth] = useState<number | null>(null)
```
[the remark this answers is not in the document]
ErikssonJoakim: The state is used to keep in memory the initial width. Without it we would have the current trigger width when performing the slideUp animation which results in a less fluid animation

src/ui/component/dropDownButton/dropDownButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type ButtonProps = Option
+type DropDownButtonProps = {
+  label: string
+  options: [Option, ...Option[]]
+  variant?: 'primary' | 'secondary'
+}
+
+const Button: FC<ButtonProps> = ({ label, onClick, icons }): JSX.Element => (
```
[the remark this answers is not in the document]
ErikssonJoakim: The button component does not entirely correspond to the buttons in this component. There are differences in border-radius, box-shadow and padding for example

src/ui/component/dragAndDrop/dragAndDrop.tsx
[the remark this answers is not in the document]
fredericvilcot: @ErikssonJoakim 
The only consideration I see here is to keep consistency through the project and provide some guidelines for our open source project.

I agree that there is no direct benefit of adding a stylesheet in terms of performance or readability, but, IMHO, mixing inline style ans stylesheets could quickly become a mess if contributors do not respect any style convention.

src/ui/component/dropDownButton/dropDownButton.tsx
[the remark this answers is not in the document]
fredericvilcot: So, it must me improved in the generic component, if not, the `Button` component has only a limited usage ;)

src/ui/component/filePicker/filePicker.tsx
ErikssonJoakim: The error handling is planned to be handled by the domain, see the ticket for selection error:
- #246

After further discussion and clarity I agree the domain should not be in a generic component, I will rework the component and add properties

src/ui/component/dropDownButton/dropDownButton.tsx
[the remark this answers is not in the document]
ErikssonJoakim: Yes makes sense I will make the necessary changes

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const handleDragAndDropState = useCallback((state: DragDropState) => {
+    state === 'dragging-over' ? setIsDraggingOver(true) : setIsDraggingOver(false)
+  }, [])
+
+  const filesInput = useMemo(
+    () =>
+      Object.assign(document.createElement('input'), {
+        onchange: onFileChange,
+        type: 'file',
+        multiple: multiple
```
fredericvilcot, line 39: Can be short-keyed 
```suggestion
        multiple
```

fredericvilcot (approved):
Great, thanks for the improvements :)

src/ui/component/filePicker/i18n/filePicker_de.json
```diff
@@ -0,0 +1,10 @@
+{
+  "file-picker": {
```
lolottetheclash, line 2: We usually use `camelCase` instead of `kebabCase` as translations keys, it could be better to harmonize.

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  return (
+    <DragAndDrop onChange={handleDragAndDropState} onDrop={onDrop}>
+      <div
+        className={classNames('okp4-dataverse-portal-file-picker-main', {
+          focused: isDraggingOver
+        })}
+      >
+        <div className="okp4-dataverse-portal-file-picker-content-container">
+          <Icon name={'folder-outlined'} />
```
lolottetheclash, line 106: you can remove `{ }`

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -0,0 +1,126 @@
...
+              {t('file-picker.drag-and-drop-files')}
+            </p>
+            {(onFileChange || onFolderChange) && (
+              <p className="okp4-dataverse-portal-file-picker-content-description">
+                {t('file-picker.or')}
+              </p>
+            )}
+          </div>
+          {(onFileChange || onFolderChange) && (
+            <div className="okp4-dataverse-portal-file-picker-button-wrapper">
```
lolottetheclash, line 118: As the display of these 2 elements depends on the same condition `(onFileChange || onFolderChange)`, is it perhaps possible to group them into one element depending on one condition?
In addition, do the "or browse" really depends on `onFileChange || onFolderChange` ?

lolottetheclash (commented):
This is a great work !!! 
I have some minor remarks & questions for you 🙂

src/ui/style/palette.scss
```diff
@@ -108,3 +108,13 @@ $tab-inactive-background-dark: rgba(255, 255, 255, 0.06);
...
 $tab-inactive-text-dark: rgba(255, 255, 255, 0.6);
 $tab-active-background: #004073;
+
+//file drag and drop
+$file-picker-background-light: #f2f6f9;
+$file-picker-background-dark: #242433;
+$file-picker-background-focused-light: #d5e5ed;
+$file-picker-background-focused-dark: #202c3f;
+$file-picker-border-light: hsla(207, 100%, 23%, 0.4);
+$file-picker-border-dark: hsl(0, 0%, 100%);
```
0xLodz, line 118: hsl for black color seems a bit overkill 😄

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -37,7 +49,14 @@ export const Collapsible: FC<CollapsibleProps> = ({
           <Icon name={iconName} />
         </div>
       </RCollapsible.Trigger>
-      <RCollapsible.Content className="okp4-dataverse-portal-collapsible-content">
+      <RCollapsible.Content
+        className={classNames('okp4-dataverse-portal-collapsible-content', contentClassName)}
+        style={{
+          // Copying the behaviour of the Radix-ui collapsible, for the slide animations
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          ['--okp4-collapsible-trigger-initial-width' as any]: `${triggerInitWidth}px`
```
0xLodz, line 57:
```suggestion
          // Copying the behaviour of the Radix-ui collapsible, for the slide animations
          ['--okp4-collapsible-trigger-initial-width' as string]: `${triggerInitWidth}px`
```

0xLodz (changes requested):
Great work! 💪

I provided some minors improvements suggestions and a question about the button.

src/ui/component/dragAndDrop/dragAndDrop.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const handleDragEnter = useCallback(() => {
+    depthRef.current++
+
+    depthRef.current === 1 && setDragDropState('dragging-over')
+  }, [])
+
+  const handleDragLeave = useCallback(() => {
+    depthRef.current--
+    depthRef.current === 0 && setDragDropState('not-dragging-over')
```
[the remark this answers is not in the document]
ErikssonJoakim: I find the state names suitable as drag-leave/enter are more events than describing a state

src/ui/component/filePicker/filePicker.tsx
ErikssonJoakim: Yes because without the callbacks there will be no other alternative than the drag and drop

src/ui/asset/icon/file-detailed.svg
```diff
@@ -0,0 +1,8 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
ErikssonJoakim: That's a good suggestion but I would prefer that we improve svg performance in a dedicated PR. That way we can optimize all in one go 🙂

src/ui/component/button/button.scss
```diff
@@ -60,6 +71,17 @@
       );
     }
 
+    &.secondary-discret {
```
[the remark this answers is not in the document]
ErikssonJoakim: This variant comes from the dropdownbutton component. Seeing as they are coupled it makes sense IMO that it is a button variant

src/ui/style/palette.scss
ErikssonJoakim: For some reason other values such as hexadecimal does not work

src/util/util.ts
```diff
@@ -35,3 +35,8 @@ export const taskLogger =
 
 export const isSubstringOf = (substring: string, source: string): boolean =>
   pipe(source, S.toLowerCase, S.includes(S.toLowerCase(substring)))
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const isError = (value: any): value is Error =>
+  value.name === 'Error' &&
+  Object.keys(value).every(key => ['cause', 'stack', 'message'].includes(key))
```
0xLodz, line 42: Checking for the name property to be exactly 'Error' is risky as it may lead to false negatives for Error subtypes.
Also, not all errors have a `cause` property, and custom error classes might not have these properties.

I would directly add the `instanceof` check within the component.

If you still want to create a typeguard function, here is my suggestion:
```suggestion
export const isError = (value: unknown): value is Error =>
  value instanceof Error;
```

src/util/util.spec.ts
```diff
@@ -49,3 +49,26 @@ describe('isSubstringOf', () => {
...
+    arg                      | expectedResult
+    ${{ name: 'Not error' }} | ${false}
+    ${{ name: 'error' }}     | ${false}
+    ${{ name: 'Error' }}     | ${false}
+    ${{}}                    | ${false}
+    ${true}                  | ${false}
+    ${false}                 | ${false}
+    ${'test string'}         | ${false}
+    ${new TypeError()}       | ${false}
+    ${new SyntaxError()}     | ${false}
```
0xLodz, line 64: Following the comment about the `isError` function implementation, I think most of these test cases are unnecessary, and the last one must provide an expected result to `true`.

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -35,24 +41,46 @@ export const FilePicker: FC<FilePickerProps> = ({
...
+      stream: file.stream()
+    }),
+    []
+  )
+
+  const handleFileChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const fileList = event.target.files
+      if (fileList?.length) {
+        const files = Array.from(fileList).map(file => toFileDTO(file))
```
0xLodz, line 81: You can simplify it :)

```suggestion
const files = Array.from(fileList, toFileDTO)
```

0xLodz (changes requested):

src/util/util.ts
ErikssonJoakim: Good point. I imagine this guard function can be useful in a general sense as catch statement arguments are unknown.

src/util/util.spec.ts
ErikssonJoakim: I think it's important to test negative cases as well though so I'd keep some test cases

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -37,7 +49,14 @@ export const Collapsible: FC<CollapsibleProps> = ({
           <Icon name={iconName} />
         </div>
       </RCollapsible.Trigger>
-      <RCollapsible.Content className="okp4-dataverse-portal-collapsible-content">
+      <RCollapsible.Content
+        className={classNames('okp4-dataverse-portal-collapsible-content', contentClassName)}
+        style={{
+          // Copying the behaviour of the Radix-ui collapsible, for the slide animations
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
0xLodz, line 56: This comment can be removed! :)

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -0,0 +1,186 @@
...
+      size: file.size,
+      mediaType: file.type,
+      stream: file.stream()
+    }),
+    []
+  )
+
+  const handleDrop = useCallback(
+    (event: React.DragEvent<HTMLDivElement>) => {
+      const files = Array.from(event.dataTransfer.items).map(item => {
```
lolottetheclash, line 58: I think you can follow same logic by passing a function as an extra arg of `Array.from` as you made in `handleFileChange`.

lolottetheclash (changes requested):
Again, great work :) 
Last remark on my side: I think you should handle the fact that when we choose a file/folder by browsing, `dropDownButton` should be closed right after, and I think it's not handled here. This behavior should be implemented in the `dropDownButton` when selecting an option.

src/ui/component/filePicker/filePicker.tsx
ErikssonJoakim: Good remarks 👍 . I've refactored the filepicker and added the closing behaviour in the dropdownbutton component.

ccamel (approved):

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const toFile = useCallback(
+    (item: DataTransferItem) => {
+      try {
+        const fsEntry = item.webkitGetAsEntry()
+        const file = item.getAsFile()
+
+        if (!fsEntry) throw new Error('Could not read item')
+        if (!file || fsEntry.isDirectory) throw new Error('Dropped item not a file')
+
+        return toFileDTO(file)
```
fredericvilcot, line 65: That's not correct to map here the data transfer File from the DOM to a `DTO` structure because this is not generic. You cannot force the user of this component to handle this file type.

I would have expected to return a `File` list (as a `DOM` type) that will be handled by the parent and mapped as its convenience.

Let's keep it like that but we have to provide the changes in a very next future.

fredericvilcot (approved):
Great work, just a structurant remark that must be addressed in a next PR, thanks 😌

Issue #240, opened 2023-05-16
[Share data] Select your data - general display 


ChaimaaZaoui:
> **As an** end-user who has selected a storage service in the in the "Select a storage service" page,
> **I want** to be able to select my dataset
> **So that** I can upload it to the storage service

### Scenario 1: general display
**Given that** I am an end user,
**And** I am using a _desktop_ device,
**When** I am in "Select a storage service" page with a selected service
**And** clicked on `Next step >`

**Then** the page is displayed as follows

* **Text**  I SHARE A DATASET
* **Figure** Stepper 
* **Text** Select your Data
* **Text** Select the dataset that you want to upload into the storage service.
* A file selection area with text and a button as follows  :
_Drag and drop your files_
Or
`Browse`
* File list with a scroll bar
* Button `< Previous step` always active
* Button `Next step >`

**When** no file is selected
**Then** the `Next step >` is disabled

### Scenario 2: File list
**Given that** I am an end user,
**And** I am using a _desktop_ device,
**When** I am in "Select your data" page with selected files
**Then** the list displays 

* File icon
* the file title as _FileTitle.Type_
* a `trash` button

**When** I click on the `trash` button
**Then** the file is deleted from the selection

### Scenario 3: Internationalisation
ENG

> Select your Data
> Select the dataset that you want to upload into the storage service.
> Drag and Drop your files
> Or
> Browse
> Previous step
> Next step

FR
> Sélectionnez vos données
> Sélectionnez le jeu de données que vous souhaitez télécharger dans le service de stockage.
> Glissez-déposez vos fichiers
> ou
> Parcourez
> Étape précédente
> Etape suivante

DE
> Wählen Sie Ihre Daten aus
> Wählen Sie den Datensatz aus, den Sie in den Speicherdienst hochladen möchten.
> Laden Sie Ihre Dateien per Drag & Drop.
> Oder
> Durchsuchen
> Vorheriger Schritt
> Nächster Schritt

#### Desktop
![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/cd3235a7-c3ce-4dfb-bced-1b7d5c0c067f)

![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/3389074c-219c-47ca-902e-a08c0c00948d)

### Environment

* staging URL: <https://app.staging.okp4.network/>

## Additional informations
None.

[...]

PR #328, opened 2023-06-26
Feat/create list and list-item components

0xLodz:
## List and ListItem components

This PR  prepares the creation of the selected data list (#240), by introducing the generic `List` and `ListItems` components, designed to provide interactive and customisable list elements.
The components are built with the intention of being easily integrated into various parts of the application that display list items, such as dataset sharing.

## Usage
```tsx
const TestListExemple: FC = () => {
  const handleClick = () => {
    console.log('clicked')
  }
  return (
    <List>
      {[1, 2, 3].map((data, i) => (
        <ListItem key={i} leftElement={<div>left element</div>} onClick={handleClick} />
      ))}

      <ListItem />
      <ListItem
        content={<div>content!</div>}
        rightElement={
          <button className="okp4-dataverse-portal-button-main">
            X
          </button>
        }
      />
    </List>
  )
}
```


src/ui/component/list/list.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const ListItem: FC<ListItemProps> = ({
+  leftElement,
+  rightElement,
+  content,
+  onClick,
+  className
+}) => (
+  <li
+    className={classNames('okp4-dataverse-portal-list-item-main', className, {
+      'pointer-cursor': onClick
```
fredericvilcot, line 22: Maybe something like `clickable` would have been more appropriated here :)

src/ui/component/list/list.scss
```diff
@@ -0,0 +1,25 @@
...
+.okp4-dataverse-portal-list-item-main {
+  @include with-theme {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding: 12px 24px;
+    border-radius: 12px;
+    background-color: themed('list-item');
+    color: themed('text');
+    font-family: 'Noto Sans', sans-serif;
```
lolottetheclash, line 19: you should use our fonts mixins ;)

src/ui/component/list/list.tsx
0xLodz: Sounds a more appropriated name 👍

lolottetheclash (changes requested):
It think you'd have to manage the space in a different way than via flex, as the elements are wrongly placed according to which you pass (see screenshot).
You'd also need to give them a bit of space between each other, because when the list item is shrunk, they're stuck together.
Capture d’écran 2023-06-27 à 09 19 06

ccamel (approved):
LGTM!

lolottetheclash (changes requested):
I think that the problem persists if you look at the screenshots here:
Capture d’écran 2023-06-27 à 12 21 37

Also, when I'm looking at the figma, content should start right after the first element:

Capture d’écran 2023-06-27 à 12 20 49

0xLodz:
@lolottetheclash Thank you for your detailed feedback! The component has been updated according to the latest mockup. As we always have content, I made it required. It reduced the number of combinations of grid template columns from 9 to 3.

src/ui/component/list/list.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const ListItem: FC<ListItemProps> = ({
+  leftElement,
+  rightElement,
+  content,
+  onClick,
+  className
+}) => {
+  console.log('render')
```
fredericvilcot, line 30: 🫣😝

0xLodz: Oops sorry! Fixed.

fredericvilcot (approved):
LGTM! 👌

src/ui/component/list/list.tsx
```diff
@@ -0,0 +1,52 @@
...
+  content: ReactElement
+  leftElement?: ReactElement
+  rightElement?: ReactElement
+  onClick?: (event: React.MouseEvent<HTMLLIElement, MouseEvent>) => void
+  className?: string
+}
+
+const gridTemplateColumnClasses = (
+  leftElement: ReactElement | undefined,
+  rightElement: ReactElement | undefined
```
lolottetheclash, line 15: you can remove `undefined` and update it with optional 😉

lolottetheclash (changes requested):
It's working well, last remarks and it will be good to go !

lolottetheclash (approved):
good for me, thanks !

PR #343, opened 2023-06-29
feat(filepicker): return file and path on file change

ErikssonJoakim:
This PR is to make the `<FilePicker />` more generic by returning the file itself on the callback. a path property is attached to simplify path handling for files without relative path.

### DIFF
--- a/src/ui/component/filePicker/filePicker.tsx
+++ b/src/ui/component/filePicker/filePicker.tsx
@@ -5,19 +5,14 @@ import { DragAndDrop } from '@/ui/component/dragAndDrop/dragAndDrop'
 import type { DragDropState } from '@/ui/component/dragAndDrop/dragAndDrop'
 import { DropDownButton } from '@/ui/component/dropDownButton/dropDownButton'
 import type { Option } from '@/ui/component/dropDownButton/dropDownButton'
+import { isNonEmpty } from 'fp-ts/lib/ReadonlyArray'
 import { Icon } from '@/ui/component/icon/icon'
 import { isError } from '@/util/util'
 import classNames from 'classnames'
 import './i18n/index'
 import './filePicker.scss'
 
-export type File = {
-  name: string
-  fullPath: string
-  size: number
-  mediaType: string
-  stream: ReadableStream<Uint8Array>
-}
+export type File = Omit<globalThis.File, 'stream'> & { stream: ReadableStream<Uint8Array> }
 
 type FilePickerProps = {
   onFileChange: (files: File[]) => void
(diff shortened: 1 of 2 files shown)


src/ui/component/filePicker/filePicker.tsx
```diff
@@ -42,14 +37,11 @@ export const FilePicker: FC<FilePickerProps> = ({
...
-      fullPath: file.webkitRelativePath,
-      size: file.size,
-      mediaType: file.type,
-      stream: file.stream()
-    }),
+  const toFileDto = useCallback(
+    (file: File): FileDTO =>
+      Object.assign(file, {
+        path: file.webkitRelativePath ? file.webkitRelativePath : `/${file.name}`
+      }),
```
fredericvilcot, line 44: If you don't mind, I prefer to not build the `path` here and let the consumer handle it from the returned `File` object; this ton respect the contract between the generic component and its caller.

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -5,22 +5,17 @@ import { DragAndDrop } from '@/ui/component/dragAndDrop/dragAndDrop'
...
-  fullPath: string
-  size: number
-  mediaType: string
-  stream: ReadableStream<Uint8Array>
-}
+export type FileDTO = File & { path: string }
 
 type FilePickerProps = {
-  onFileChange: (files: File[]) => void
+  onFileChange: (files: FileDTO[]) => void
```
fredericvilcot, line 18: Let's handle here the DOM `File` type ;)

src/ui/component/filePicker/filePicker.tsx
```diff
@@ -62,32 +54,36 @@ export const FilePicker: FC<FilePickerProps> = ({
...
+        return toFileDto(file)
       } catch (error: unknown) {
         isError(error) && onError?.(error)
       }
     },
-    [onError, toFileDTO]
+    [onError, toFileDto]
+  )
+
+  const isFileDtoArray = useCallback(
```
fredericvilcot, line 65: What's the use case in which a file could be `undefined` into the array?

fredericvilcot (changes requested):

src/ui/component/filePicker/filePicker.tsx
ErikssonJoakim: Sure I'll make the necessary changes :) It makes more sense anyway

src/ui/component/filePicker/filePicker.tsx
ErikssonJoakim: We can drop anything draggable item in the drop zone so we might end up having a variety of non-file elements.

0xLodz (approved):
Good job! 🚀 

Considering future enhancements, it might be more efficient to handle file streams via Blob URLs. This approach would let us extract File data directly from the Blob URL, preventing the need to store the stream within our Zustand state.

An issue will be created to explore this potential enhancement.

fredericvilcot (approved):
Great, thx :)

PR #352, opened 2023-06-30
Feat/add feature not available layout

lolottetheclash:
This PR allows us to add a shared layout for our app routes.
Here, the `share` part is unavailable for small screens, so we render an "Unsupported Feature" if the user uses one.


AnjaMis (approved):

fredericvilcot (approved):
Great! 😊

PR #354, opened 2023-06-30
Feat/add list item transition

lolottetheclash:
This PR adds the list item transition (visible in data selection step) by updating the component `List` with `React CSS Transition`.

For now the transition is 400ms to be compliant will all portal transitions (to be configured in the future)


ccamel (approved):
LGTM!

PR #362, opened 2023-07-03
Feat/file errors

fredericvilcot:
ℹ️ This PR handles the possible errors returned by the `storeFiles` and `removeFile` commands by:

- logging them to the console for debugging purpose (dev mode)
- displaying a generic snackbar for the user

👉 Implements #246

### DIFF
--- a/src/domain/file/command.ts
+++ b/src/domain/file/command.ts
@@ -16,26 +16,26 @@ export type StoreFileInput = {
 
 export type StoreFilesInput = StoreFileInput[]
 
-export const ResourceConflictError = (fileIds: FileId[]) =>
+export const ResourceConflictError = (resourceIds: FileId[]) =>
   ({
     _tag: 'resource-conflict',
-    fileIds
+    resourceIds
   } as const)
 
 /**
- *  Error thrown when an id of a file to be stored already exists in memory,
+ *  Error when an id of a file to be stored already exists in memory,
  *  or if multiple files in the StoreFileInput payload have the same id
  */
 export type ResourceConflictError = ReturnType<typeof ResourceConflictError>
 
-export const ResourceNotFoundError = (fileId: FileId) =>
+export const ResourceNotFoundError = (resourceId: FileId) =>
   ({
     _tag: 'resource-not-found',
-    fileId
+    resourceId
   } as const)
 
 /**
- *  Error thrown when the id of a file to be removed is not found in the list of stored files in memory.
+ *  Error when the id of a file to be removed is not found in the list of stored files in memory.
  */
 export type ResourceNotFoundError = ReturnType<typeof ResourceNotFoundError>
 
@@ -52,10 +52,12 @@ export const ShowFileError: Show<FileError> = {
   show: (error: FileError): string => {
     switch (error._tag) {
       case 'resource-conflict': {
-        return `Oops... You are trying either to store a file whose id already exists in memory or to store files with the same id... So we can't store these files with these ids: [${error.fileIds}]`
+        return `Error ${
+          error._tag
+        }: Failed to store resource with conflicting IDs [${error.resourceIds.join(', ')}].`
       }
       case 'resource-not-found': {
-        return `Oops... The provided id '${error.fileId}' does not exist in memory... So we can't remove this file...`
+        return `Error ${error._tag}: Failed to remove resource with ID '${error.resourceId}' since it does not exist.`
       }
     }
   }
--- a/src/domain/file/command.ts
+++ b/src/domain/file/command.ts
@@ -1,4 +1,5 @@
 import type { IOEither } from 'fp-ts/IOEither'
+import type { Show } from 'fp-ts/lib/Show'
 
 export type MediaType = string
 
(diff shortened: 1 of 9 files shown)


src/ui/page/share/dataset/steps/dataSelection/dataSelection.tsx
```diff
@@ -7,8 +9,36 @@ import { FilePicker } from '@/ui/component/filePicker/filePicker'
...
+  type DispatchNotificationInput
+} from '@/ui/hook/useDispatchNotification'
+
+type FileError = ResourceConflictError | ResourceNotFoundError
+
+const logFileError = (error: FileError): void => {
+  switch (error._tag) {
+    case 'resource-conflict': {
+      console.error(
+        "Oops... You are trying either to store a file whose id already exists in memory or to store files with the same id... So we can't store these files..."
```
ccamel, line 30: I'm a little latle, but from what I see the resource identifiers in conflict are missing in the error definition, which is unfortunate. So the reported log therefore lacks of context, imho.

ccamel, line 30: The most canonical way would be to use the [Show](https://gcanti.github.io/fp-ts/modules/Show.ts.html#show-overview) class type, and have it implemented by errors. Then, to have an IO for logging (but you can always use an effectful function if you prefer)

ccamel (changes requested):

src/domain/file/command.ts
```diff
@@ -15,9 +16,10 @@ export type StoreFileInput = {
 
 export type StoreFilesInput = StoreFileInput[]
 
-export const ResourceConflictError = () =>
+export const ResourceConflictError = (fileIds: FileId[]) =>
```
ccamel, line 19: If the error is called `ResourceConflictError`, then the argument and properties should use the naming `resource` and not `file`.

src/domain/file/domain.ts
```diff
@@ -37,7 +37,8 @@ const eqFile: Eq<{ id: string }> = pipe(
   eqContramap(it => it.id)
 )
 
-const throwResourceConflictError = (): ResourceConflictError => ResourceConflictError()
+const throwResourceConflictError = (filesIds: FileId[]): ResourceConflictError =>
```
ccamel, line 40: The name is misleading because it suggests that the function throws an error. But you don't throw anything here. The function is actually a constructor. So I'd rename it: `createResourceConflictError` or `makeResourceConflictError` or simpler `resourceConflictError`.

src/domain/file/command.ts
```diff
@@ -43,3 +45,18 @@ export type Command = {
...
   removeFile: (fileId: FileId) => IOEither<ResourceNotFoundError, void>
 }
+
+export type FileError = ResourceConflictError | ResourceNotFoundError
+
+export const ShowFileError: Show<FileError> = {
+  show: (error: FileError): string => {
+    switch (error._tag) {
+      case 'resource-conflict': {
+        return `Oops... You are trying either to store a file whose id already exists in memory or to store files with the same id... So we can't store these files with these ids: [${error.fileIds}]`
```
ccamel, line 55: I'd be inclined to make a more summary error message:

```suggestion
        return `Error ${error._tag}: Failed to store resource with conflicting IDs [${error.fileIds.join(', ')}].`
```

If you want to improve the message, it's up to the caller to do so, notably by providing a little more context according to the use case.

src/domain/file/command.ts
```diff
@@ -43,3 +45,18 @@ export type Command = {
...
+      }
+      case 'resource-not-found': {
+        return `Oops... The provided id '${error.fileId}' does not exist in memory... So we can't remove this file...`
```
ccamel, line 58:
```suggestion
        return `Error ${error._tag}: Failed to remove resource with ID '${error.fileId}' since it does not exist.`
```

src/ui/page/share/dataset/steps/dataSelection/dataSelection.tsx
```diff
@@ -17,6 +28,21 @@ export const DataSelection: FC = () => {
...
+
+  const handleFileError = useCallback(
+    (error: FileError) => {
+      const notificationInput: DispatchNotificationInput = {
+        type: 'error',
+        titleKey: 'error.problem',
+        messageKey: 'error.processing',
+        action: 'refresh'
+      }
+      console.error(ShowFileError.show(error))
```
ccamel, line 41: You can wrap the error message here to bring more context if you want, or to make the logged message more friendly.

ccamel (changes requested):

src/domain/file/command.ts
fredericvilcot: Really much better 👍

src/domain/file/command.ts
fredericvilcot: Really much better 👍

src/ui/page/share/dataset/steps/dataSelection/dataSelection.tsx
fredericvilcot: @ccamel 
AFAIS, the error messages returned by the `show` function look pretty standalone as they contain the whole context related to the use-case.

So, I'm not sure here we need to provide additional context, so, if you are OK, I'll suggest to keep it like it is.

0xLodz (approved):
LGTM!

Issue #345, opened 2023-06-29
[Portal] Button behaviour


ChaimaaZaoui:
As a user, I want a button with four states (Default, Hover, Pressed, and Disabled) so that I can interact with it effectively.

### Scenario 1: Button behaviour

Given a button with four states (Default, Hover, Pressed, and Disabled),

When the button is in the Default state, 
Then I can see the label and it appears clickable.

When I hover over the button
Then it transitions to the Hover state as designed.

**When I click on the button while in the Hover state, 
Then it transitions to the Pressed state the few micro seconds until the action behind the button is executed.
And If I release the mouse button while in the Pressed state, the button returns to the Default state.** 

When in the Disabled state, 
Then the button's appearance indicates it cannot be interacted with, and clicking is not possible.

_This is applicable to buttons, and to cards that have a button behaviour
Please go through the whole application to make sure that all buttons have the 4 states_

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206525668) by [Unito](https://www.unito.io)

fredericvilcot:
For now, it only misses the `pressed` state that must be implemented in the generic `Button` component.

ChaimaaZaoui:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/DEV-board--63e210cc9b1f2f73ee3f4e5f/issues/okp4/dataverse-portal/345?planning-poker) with Zenhub @0xLodz @AnjaMis @ErikssonJoakim @fredericvilcot

ChaimaaZaoui:
Tested on desktop => Success
Tested on mobile => Success

PR #367, opened 2023-07-06
feat: add button pressed state

AnjaMis:
This PR adds a button pressed state to the Button component.

Issue #345


src/ui/component/button/button.scss
```diff
@@ -18,6 +18,10 @@
     box-shadow: none;
   }
 
+  &:focus {
+    background-color: $pressed-color;
+  }
```
0xLodz, line 23: Shadow can also be removed on focus!
```suggestion
  &:focus {
    box-shadow: none;
    background-color: $pressed-color;
  }
```

src/ui/component/card/card.scss
```diff
@@ -24,6 +24,10 @@ $backgrounds: (
           background-color: themed(card-background-hover);
         }
 
+        &:focus {
+          background-color: themed('card-background-pressed');
+        }
```
0xLodz, line 29: I am probably missing something, because during testing I was not able to focus on the card, but only the child button. We might want to use `:focus-within` here instead of `:focus`. It tried with `focus-within`, It worked for me but only outside of the background loop.

0xLodz (changes requested):

src/ui/component/card/card.scss
0xLodz: Sorry for misleading you, your implementation is correct because we must only target interact cards. Since these cards are actually `div` element, they are not focusable by a user. We might create an issue in order to make them accessible by either transforming these divs in `button` tag or providing them a `tabindex`.

0xLodz (approved):
LGTM!

src/ui/component/button/button.scss
AnjaMis: done, thanks!

fredericvilcot (approved):
👍

PR #371, opened 2023-07-06
fix(infra): escape double quotes in filter literals

fredericvilcot:
This PR adds escaping for double quotes in the `FILTER` operator for dynamic literals that lead to a 400 http error.

### DIFF
--- a/src/infra/dataverse/sparql/sparqlGateway.ts
+++ b/src/infra/dataverse/sparql/sparqlGateway.ts
@@ -10,7 +10,7 @@ import type {
   RetrieveDataverseResult,
   DataverseElementType
 } from '@/domain/dataverse/port'
-import { escapeRegExp, getURILastElement } from '@/util/util'
+import { escapeSparqlStr, getURILastElement } from '@/util/util'
 import { createAbortableFetch } from '@/util/fetch/fetch'
 import type { SparqlBinding, SparqlResult } from './dto'
 const { abortRequest, fetchWithAbort } = createAbortableFetch()
@@ -32,7 +32,7 @@ export const sparqlGateway: DataversePort = {
         byProperty,
         O.map(
           filter =>
-            `FILTER ( contains(lcase(str(?${filter.property})), "${escapeRegExp(
+            `FILTER ( contains(lcase(str(?${filter.property})), "${escapeSparqlStr(
               filter.value.toLowerCase()
             )}" ) )`
         ),
--- a/src/infra/dataverse/sparql/sparqlGateway.ts
+++ b/src/infra/dataverse/sparql/sparqlGateway.ts
@@ -30,13 +30,13 @@ export const sparqlGateway: DataversePort = {
     const byPropertyFilter = (): string =>
       pipe(
         byProperty,
-        O.map(
+        O.match(
+          () => '',
           filter =>
             `FILTER ( contains(lcase(str(?${filter.property})), "${escapeSparqlStr(
               filter.value.toLowerCase()
             )}" ) )`
-        ),
-        O.getOrElse(() => '')
+        )
       )
 
     const byServiceCategoryFilter = (): string =>
--- a/src/infra/dataverse/sparql/sparqlGateway.ts
+++ b/src/infra/dataverse/sparql/sparqlGateway.ts
@@ -10,7 +10,7 @@ import type {
   RetrieveDataverseResult,
   DataverseElementType
 } from '@/domain/dataverse/port'
-import { getURILastElement } from '@/util/util'
+import { escapeRegExp, getURILastElement } from '@/util/util'
 import { createAbortableFetch } from '@/util/fetch/fetch'
 import type { SparqlBinding, SparqlResult } from './dto'
 const { abortRequest, fetchWithAbort } = createAbortableFetch()
@@ -32,9 +32,9 @@ export const sparqlGateway: DataversePort = {
         byProperty,
         O.map(
           filter =>
-            `FILTER ( contains(lcase(str(?${filter.property})), "${filter.value
-              .toLowerCase()
-              .replace(/"/g, '\\"')}" ) )`
+            `FILTER ( contains(lcase(str(?${filter.property})), "${escapeRegExp(
+              filter.value.toLowerCase()
+            )}" ) )`
         ),
         O.getOrElse(() => '')
       )
--- a/src/infra/dataverse/sparql/sparqlGateway.ts
+++ b/src/infra/dataverse/sparql/sparqlGateway.ts
@@ -32,7 +32,9 @@ export const sparqlGateway: DataversePort = {
         byProperty,
         O.map(
           filter =>
-            `FILTER ( contains(lcase(str(?${filter.property})), "${filter.value.toLowerCase()}" ) )`
+            `FILTER ( contains(lcase(str(?${filter.property})), "${filter.value
+              .toLowerCase()
+              .replace(/"/g, '\\"')}" ) )`
         ),
         O.getOrElse(() => '')
       )
--- a/src/util/util.spec.ts
+++ b/src/util/util.spec.ts
@@ -75,26 +75,29 @@ describe('isError guard function', () => {
 
 describe('escapeSparqlStr', () => {
   describe.each`
-    arg             | expectedResult
-    ${undefined}    | ${''}
-    ${'\t'}         | ${'\\t'}
-    ${'\n'}         | ${'\\n'}
-    ${'\r'}         | ${'\\r'}
-    ${'\b'}         | ${'\\b'}
-    ${'\f'}         | ${'\\f'}
-    ${'"'}          | ${'\\"'}
-    ${"'"}          | ${"\\'"}
-    ${'\\'}         | ${'\\\\'}
-    ${'foo\nbar'}   | ${'foo\\nbar'}
-    ${'foo\n'}      | ${'foo\\n'}
-    ${'\nfoo'}      | ${'\\nfoo'}
-    ${'foo"bar'}    | ${'foo\\"bar'}
-    ${"foo'bar"}    | ${"foo\\'bar"}
-    ${'foo\\bar'}   | ${'foo\\\\bar'}
-    ${'\\u00a0'}    | ${'\\\\u00a0'}
-    ${'\u00a0'}     | ${' '}
-    ${'\u005Cbar'}  | ${'\\\\bar'}
-    ${'\\u005Cbar'} | ${'\\\\u005Cbar'}
+    arg                    | expectedResult
+    ${undefined}           | ${''}
+    ${'\t'}                | ${'\\t'}
+    ${'\n'}                | ${'\\n'}
+    ${'\r'}                | ${'\\r'}
+    ${'\b'}                | ${'\\b'}
+    ${'\f'}                | ${'\\f'}
+    ${'"'}                 | ${'\\"'}
+    ${"'"}                 | ${"\\'"}
+    ${'\\'}                | ${'\\\\'}
+    ${'foo\nbar'}          | ${'foo\\nbar'}
+    ${'foo\n'}             | ${'foo\\n'}
+    ${'\nfoo'}             | ${'\\nfoo'}
+    ${'foo"bar'}           | ${'foo\\"bar'}
+    ${"foo'bar"}           | ${"foo\\'bar"}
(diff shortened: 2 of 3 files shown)


0xLodz (approved):
👍

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -29,7 +29,9 @@ export const sparqlGateway: DataversePort = {
       flow(A.map(buildStrExpression), a => a.join(' || '))(filters)
 
     const byPropertyFilter = (filter: ByPropertyQueryFilter): string =>
-      `contains(lcase(str(?${filter?.property})), "${filter?.value.toLowerCase()}" )`
+      `contains(lcase(str(?${filter?.property})), "${filter?.value
+        .toLowerCase()
+        .replace(/"/g, '\\"')}" )`
```
ccamel, line 34: Not quite right... What you want is to _escape_ in order to safely include the content within a SPARQL string. For this, you have to consider all the cases, for instance the character `\`.

Because this is a more general concern to convert literal text into a format safe for inclusion in a particular context,  I'd recommand to have dedicated functions for that.

ccamel (changes requested):

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: @ccamel 
Your are completely right.
I added a simple fix because this was the only character that was not correctly handled in the `sparql` query.
I pushed a more generic function about escaping, tell me what do you think 😌

src/util/util.ts
```diff
@@ -37,3 +37,9 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
   pipe(source, S.toLowerCase, S.includes(S.toLowerCase(substring)))
 
 export const isError = (value: unknown): value is Error => value instanceof Error
+
+export const escapeRegExp = (str?: string): string => {
+  const regExp = /[\\^"'$.*+?()[\]{}|]/g
+  const regExpFromSource = RegExp(regExp.source)
```
ccamel, line 43: Seems redundant because `regExp` is already a RegExp object.

src/util/util.ts
```diff
@@ -37,3 +37,9 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
   pipe(source, S.toLowerCase, S.includes(S.toLowerCase(substring)))
 
 export const isError = (value: unknown): value is Error => value instanceof Error
+
+export const escapeRegExp = (str?: string): string => {
```
ccamel, line 41: I wouldn't have called the function like this because it suggests that the escaping is to put safely a string into a regexp, which is not the case (even if the escaped symbols overlap). Instead, I would have called thefunction:  `escapeSparqlStr`.

src/util/util.ts
```diff
@@ -37,3 +37,9 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
...
+  const regExp = /[\\^"'$.*+?()[\]{}|]/g
```
ccamel, line 42: Did you test it on a SPARQL query ? Escaping `$`, `.` and so on will not work.

> FILTER ( contains(lcase(str(?title)), "\." ) )
>  Encountered: "." (46), after : "\"\\"

Cf. https://www.w3.org/TR/sparql11-query/#grammarEscapes

ccamel (changes requested):

src/util/util.ts
0xLodz: @ccamel thanks for the specification link.  We probably only need to manage the escape sequence provided by the Sparql spec.
The function would be:
```ts
export const escapeSparqlStr = (str?: string): string => {
  const sparqlEscapeRegExp = /[\t\n\r\b\f"\\']/g
  const escapeCharsMap: { [key: string]: string } = {
    '\t': '\\t',
    '\n': '\\n',
    '\r': '\\r',
    '\b': '\\b',
    '\f': '\\f',
    '"': '\\"',
    `'`: `\\'`,
    '\\': '\\\\',
  }

  return str ? str.replace(sparqlEscapeRegExp, (char) => escapeCharsMap[char]) : '';
}
```
It would be easier to read if we use a Map or something similar because we don't have that many characters to escape and the replacement string doesn't follow the same pattern. So using the entire match (`$&`) with a leading slash doesn't cover all cases.
I used backticks in the exemple for escaping single quote for readability.

src/util/util.spec.ts
```diff
@@ -72,3 +72,32 @@ describe('isError guard function', () => {
     })
   })
 })
+
+describe('escapeSparqlStr', () => {
```
ccamel, line 76: Great to have tests. :+1:

However, they don't cover certain cases. For example, the tests assert the escape of one single character, but not several characters in a string (which is important to ensure). On the other hand, it does not proove that certain characters that might be considered special are not escaped, which is intended to be compiant with the w3c spec.

So I'd suggest extending the test case coverage

src/util/util.ts
```diff
@@ -37,3 +37,20 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
   pipe(source, S.toLowerCase, S.includes(S.toLowerCase(substring)))
 
 export const isError = (value: unknown): value is Error => value instanceof Error
+
+export const escapeSparqlStr = (str?: string): string => {
+  const sparqlEscapeRegExp = /[\t\n\r\b\f"'\\]/g
+
+  const escapeCharactersMap = new Map([
```
ccamel, line 44: why Map? you can use a simple object here (more performent and convenient).

src/util/util.ts
```diff
@@ -37,3 +37,20 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
   pipe(source, S.toLowerCase, S.includes(S.toLowerCase(substring)))
 
 export const isError = (value: unknown): value is Error => value instanceof Error
+
+export const escapeSparqlStr = (str?: string): string => {
+  const sparqlEscapeRegExp = /[\t\n\r\b\f"'\\]/g
```
ccamel, line 42: I think you can avoid the use of the regexp here, if you have the mapping of escaped characters. It will be more efficient, and more maintenable, imho.

ccamel (changes requested):

src/util/util.spec.ts
0xLodz: Didn't think of these cases, thanks! They were addressed in [3bcb99c](https://github.com/okp4/dataverse-portal/pull/371/commits/3bcb99cf3ea1e8722f117175071573cea29bfb29).

src/util/util.ts
0xLodz: The function has been refactored according to you feedbacks in [05745c0](https://github.com/okp4/dataverse-portal/pull/371/commits/05745c05fd013b3864ec2d203d0c22c4ef3792e1). Thanks

src/util/util.ts
```diff
@@ -37,3 +37,30 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
...
+    '\f': '\\f',
+    '"': '\\"',
+    "'": "\\'",
+    '\\': '\\\\'
+  }
+
+  return pipe(
+    str,
+    O.fromNullable,
+    O.map(s =>
```
ccamel, line 56: Use `O.match` for final unboxing (as said above).

src/util/util.ts
```diff
@@ -37,3 +37,30 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
...
+      pipe(
+        s,
+        Array.from,
```
ccamel, line 59: Use `A.of` instead. This will help resolve the type inference you had.

src/util/util.ts
```diff
@@ -37,3 +37,30 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
...
+        A.map((char: string) => escapeCharactersMap[char] || char),
+        (chars: string[]) => chars.join('')
```
ccamel, line 61: Remove type annotation and let TypeScript infer types.

src/util/util.ts
```diff
@@ -37,3 +37,30 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
...
+    '"': '\\"',
+    "'": "\\'",
+    '\\': '\\\\'
+  }
+
+  return pipe(
+    str,
+    O.fromNullable,
+    O.map(s =>
+      pipe(
```
ccamel, line 57: Leverage `flow` operator to take advantage  [η-reduction](https://sookocheff.com/post/fp/eta-conversion/).

src/util/util.ts
```diff
@@ -37,3 +37,30 @@ export const isSubstringOf = (substring: string, source: string): boolean =>
...
+        s,
+        Array.from,
+        A.map((char: string) => escapeCharactersMap[char] || char),
```
ccamel, line 60: Use [R.lookup](https://gcanti.github.io/fp-ts/modules/Record.ts.html#lookup) instead.

ccamel (changes requested):

src/util/util.ts
0xLodz: I tried to apply all your recommendations, and it worked pretty well except for this one, I ran into some typing issues. The `A.of` function takes a value and puts it into an array, whereas `Array.from` converts an iterable (or array-like) object into an array. Since I needed to convert a string into an array of its characters, I didn't find a cleaner solution than `Array.from`.

Is there a way to use `A.of` in this case? Thanks

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -32,7 +32,9 @@ export const sparqlGateway: DataversePort = {
         byProperty,
         O.map(
           filter =>
-            `FILTER ( contains(lcase(str(?${filter.property})), "${filter.value.toLowerCase()}" ) )`
+            `FILTER ( contains(lcase(str(?${filter.property})), "${escapeSparqlStr(
```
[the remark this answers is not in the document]
0xLodz: That's definitely easier to read! Thanks

src/util/util.ts
0xLodz: Thanks for the link!

src/util/util.ts
0xLodz: Neat!

src/util/util.ts
ccamel: Ah, my bad. I see your point. `A.of just` creates an array with the string as unique element. It could be done in pure fp-ts, but it adds some boilerplate code:

```ts
pipe(
  s,
  S.split(''),
  A.of
)
```

If you want to continue down this path, you'll need to add a utility function that encapsulates this code. But I'll let you decide.

0xLodz: Using `S.split('')` seemed more natural to me, but we need to deal with read-only Array.
And my bad, I forgot about using directly `ReadonlyArray` with `map`.
Latest commit was updated by and `Array.from` was replaced with `S.split('')`.
```ts
S.split(''),
RA.map(c =>
```

ccamel: You can use [`S.split('')`](https://gcanti.github.io/fp-ts/modules/string.ts.html#split) which returns an `ReadonlyNonEmptyArray`  of characters. But it's not that impressive.

```ts
  pipe(
    s,
    S.split(''),
  )
```

if I combine everything together, we should get something like this:

```ts
export const escapeSparqlStr = (str?: string): string => {
  const escapeCharactersMap: Record<string, string> = {
    '\t': '\\t',
    '\n': '\\n',
    '\r': '\\r',
    '\b': '\\b',
    '\f': '\\f',
    '"': '\\"',
    "'": "\\'",
    '\\': '\\\\'
  }

  const lookupEscaperCharacter = (c: string): string =>
    pipe(
      escapeCharactersMap,
      R.lookup(c),
      O.getOrElse(constant(c))
    )

  return pipe(
    str,
    O.fromNullable,
    O.getOrElse(constant(S.empty)),
    S.split(S.empty),
    RA.map(lookupEscaperCharacter),
    RA.reduce(S.Monoid.empty, S.Monoid.concat)
  )
}
```

There must be room for improvement, but I'm running out of time. I leave this in your hands for integration, at your convenience.

0xLodz: That's a fantastic way to make it easier to read and understand. We can use your solution. Thanks!

ccamel (approved):
Let's go! :+1:

fredericvilcot (commented):
Great, thank you both @0xLodz & @ccamel 👍

lolottetheclash (approved):
Great Job!!! 👍

PR #373, opened 2023-07-07
Feat/create Field component

0xLodz:
# Field component creation

This PR introduces the `Field` component, a reusable, controlled component designed to provide an interactive and customisable text input, number input or textarea field. The component is built with the intention of being integrated into various parts of the application that require form functionality, such as Share Data step 3.

### Usage
The `onChange` function prop that gets triggered whenever the input value changes.

In the following example, `textareaValue` and `textInputValue` are the current fields values controlled by the parent component, and `handleChangeInput` is a function that updates the state of the parent component when the input value changes.

```tsx
const [error, setError] = useState<string>('sds')
const [textareaValue, setTextareaValue] = useState<string>('')

const handleChangeInput = useCallback((value: string): void => {
  setTextareaValue(value)
  setError('')
}, [])

<Field
  error={error}
  id="3"
  label="I am a label"
  onChange={handleChangeInput}
  rightElement={<p>KNOW</p>}
  value={textareaValue}
/>
<Field
  disabled
  error={error}
  id="1"
  label="test"
  multiline
  onChange={handleChangeInput}
  required
  resizable
  value={textareaValue}
/>
```

#### Inputs state
image

#### Textarea
image

image

image

#### Input text (read only / disabled state)
image

### ⚠️ Note
Inputs with placeholders and optional label (e. g. screenshot below) will be handled in the next iteration. It requires some extra style rules and are not necessary for the Share data step 3.
image


src/ui/component/textField/textField.tsx
```diff
@@ -0,0 +1,98 @@
...
+      {multiline ? (
+        <textarea
+          className={inputClassName}
+          disabled={disabled}
+          id={id}
+          name={id}
+          onChange={handleChange}
+          readOnly={readonly}
+          required={required}
+          value={value}
```
fredericvilcot, line 66: You can prevent from duplicating the same props between `textarea` and `input` by using `React.cloneElement`

src/ui/component/textField/textField.tsx
```diff
@@ -0,0 +1,98 @@
...
+          type="text"
+          value={value}
+        />
+      )}
+      {icons?.startIcon && (
+        <div className="okp4-dataverse-portal-text-field-start-icon">{icons.startIcon}</div>
+      )}
+      {icons?.endIcon && (
+        <div className="okp4-dataverse-portal-text-field-end-icon">{icons.endIcon}</div>
+      )}
```
fredericvilcot, line 86: What will happen when `multiline` / `resizable` props are combined with icons?
Are the `icons` rendered correctly?

src/ui/component/textField/textField.tsx
0xLodz: Didn't find a way to prevent duplication using `cloneElement` since these elements represents different html tags. Instead, the duplicated elements attributes has been set in an object that is spread at JSX level to prevent duplication.

Is this an appropriate solution ?

src/ui/component/textField/textField.tsx
0xLodz: Icons are located at the top of the input, either with or without `multiline` and `resizable` properties.

src/ui/component/field/field.tsx
```diff
@@ -0,0 +1,108 @@
+import { type FC, type ChangeEvent, useCallback, useState, useRef } from 'react'
+import classNames from 'classnames'
+import { Icon } from '@/ui/component/icon/icon'
+import './field.scss'
+
+type FieldProps = {
+  id: string
+  label: string
+  onChange: (value: string) => void
+  value?: string
```
fredericvilcot, line 10: If `type = 'number'`, value's type should be `number`, no?

src/ui/component/field/field.tsx
```diff
@@ -0,0 +1,108 @@
...
+          className={classNames(textFieldClassNames, 'okp4-dataverse-portal-field-left-element')}
+        >
+          {leftElement}
+        </div>
+      )}
+
+      {multiline ? (
+        <textarea {...textFieldInputProps} />
+      ) : (
+        <input {...textFieldInputProps} type={type === 'number' ? 'number' : 'text'} />
```
fredericvilcot, line 84: Seems unnecessary as `type` property as a default value

src/ui/component/field/field.tsx
0xLodz: HTML input value are always strings, even this may seem unintuitive for input `number` type 😅

src/ui/component/field/field.tsx
0xLodz: Oops! Thank you

src/ui/component/field/field.tsx
fredericvilcot: 🤪

fredericvilcot (approved):
Great, thx!

PR #385, opened 2023-07-11
Feat/share data slice

fredericvilcot:
👉 This PR adds a `ShareDataFormSlice` in the current `appStore` that will handle transient states for the share data form.

ℹ️ This slice handles a specific model (entity) that will be used to validate the form in the future business domain and as the data source for the request in last step.

🧪 The slice is fully tested and serve as documentation

⚠️ ⚠️ Errors should be handled for commands, in particular to check the resource ids. For the sake of speed, they will be managed in a second time. ⚠️⚠️


lolottetheclash (approved):
Great ! 🚀

PR #386, opened 2023-07-12
feat/dropdown-component

oussama-aitbaziz-dev:
# DropDown 
This PR aims to add  the multiselect variant of `<DropDown />` component, other variants and UI behaviors will be implemented later.

### Renders
Capture d’écran 2023-07-18 à 23 57 44Capture d’écran 2023-07-18 à 23 58 11

Capture d’écran 2023-07-18 à 23 57 50Capture d’écran 2023-07-18 à 23 58 07

### Behaviour
The component gives users the ability to select multiple options from a given list of options
 
### Usage
```
const [value, setValue] = useState([])

 <DropDown
      onChange={values => setValue(values)}
      options={['item 1', 'item 2', 'item 3', 'item 4', 'item 5', 'item 6', 'item 7']}
      placeholder="Placeholder"
      searchPlaceholder="Search"
      selectionType="checkbox"
      value={value}
    />
```


src/ui/component/dropDown/dropDown.tsx
```diff
@@ -1,168 +1,152 @@
...
+  filterValues,
+  searchPlaceholder,
+  selectionType,
+  maxSearchResults = 10
+}) => {
+  const createDefaultFilterOption = (value: string): SelectionOption => ({
+    name: filterName,
+    value,
+    selected: false,
+    disabled: false
```
fredericvilcot, line 40: These properties should be optional, IMHO It's better to initialize them to `false`

src/ui/component/dropDown/dropDown.tsx
```diff
@@ -1,168 +1,152 @@
...
-
-  const selectOption = useCallback(
-    (optionValue: string, subOptionValue?: string) => {
-      onChange(optionValue, subOptionValue)
-      toggleMenu()
-    },
-    [onChange, toggleMenu]
+export const DropDown: FC<DropDownProps> = ({
+  filterName,
+  filterValues,
```
fredericvilcot, line 31: Here you want to control the component by giving back the selected options, so maybe you could handle directly a list of them, that looks much more explicit, IMHO.

fredericvilcot (changes requested):

src/ui/component/dropDown/dropDown.tsx
oussama-aitbaziz-dev: I'm not sure what you mean by that, they are already initialized to false, its a reprensatation of their state

fredericvilcot: I mean that, if they are optional properties, they are evaluated to false by default ;)

src/ui/component/dropDown/dropDown.tsx
```diff
@@ -1,168 +1,152 @@
...
+          </div>
+        }
+        iconName="chevron-sharp"
+        open
+        trigger={
+          <div className="okp4-dataverse-portal-dropdown-header">
+            {selectedOptions.length ? (
+              <div className="okp4-dataverse-portal-dropdown-header-selection">
+                {selectedOptions.slice(0, TAGS_DISPLAY_LIMIT).map(({ value, name }) => (
+                  <Tag
```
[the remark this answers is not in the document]
oussama-aitbaziz-dev: If I understand correctly your suggestion will imply that all state management currently done by the DropDown component to be done by an eventual parent component that will render it ??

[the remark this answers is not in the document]
fredericvilcot: Exactly.
If not, this not a generic component

src/ui/component/dropDown/dropDown.tsx
```diff
@@ -1,168 +1,152 @@
...
-      onChange(optionValue, subOptionValue)
-      toggleMenu()
-    },
-    [onChange, toggleMenu]
+export const DropDown: FC<DropDownProps> = ({
+  filterName,
+  filterValues,
+  searchPlaceholder,
+  selectionType,
+  maxSearchResults = 10
```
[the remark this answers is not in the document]
oussama-aitbaziz-dev: Is this a mistake? I see that `maxSearchResults` is already parametrized

src/ui/component/dropDown/dropDown.tsx
oussama-aitbaziz-dev: I see!

src/ui/component/tag/tag.tsx
```diff
@@ -2,17 +2,19 @@ import { type FC } from 'react'
 import { Icon } from '@/ui/component/icon/icon'
 import './tag.scss'
 import { useTranslation } from 'react-i18next'
+import classNames from 'classnames'
 
 type TagProps = {
   tagName: string
+  className?: string
```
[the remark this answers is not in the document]
oussama-aitbaziz-dev: I think think of a better way to overwrite styles that doesnt imply using specific classes

src/ui/component/dropDown/dropDown.tsx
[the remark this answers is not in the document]
fredericvilcot: I mean you should use an environment variable to easily control this parameter.

src/ui/component/dropDown/dropDown.tsx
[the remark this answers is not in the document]
fredericvilcot: OK, we decided to make a specific component adapted to the view for now.
A generic component will come in a next future.

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,157 @@
...
+}
+
+type SelectionItemType = 'checkbox'
+
+type DropDownProps = {
+  filterName: string
+  filterValues: string[]
+  searchPlaceholder: string
+  selectionType: SelectionItemType
+  maxSearchResults?: number
```
fredericvilcot, line 26: Should not be a props if we consider that it's an environment variable.

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,157 @@
...
+  onSelectionChange: (values: string[]) => string[]
```
fredericvilcot, line 27: Callback function should return nothing.
```suggestion
  onSelectionChange: (value: string[]) => void
```

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,157 @@
...
+    value
+  })
+
+  const [searchTerm, setSearchTerm] = useState<string>('')
+
+  const [filterOptions, setFilterOptions] = useState<DropDownOption[]>(
+    filterValues
+      .map(value => createDefaultFilterOption(value))
+      .sort((a, b) => a.value.localeCompare(b.value))
+  )
```
fredericvilcot, line 50: If the component is purely specific and controlled by its parent, you don't need a local state.

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const selectedOptions = useMemo(
+    () => filterOptions.filter(({ selected }) => selected),
+    [filterOptions]
+  )
+
+  useEffect(() => {
+    if (selectedOptions.length) {
+      onSelectionChange(selectedOptions.map(({ value }) => value))
+    }
+  }, [onSelectionChange, selectedOptions])
```
fredericvilcot, line 91: Useless, `onSelectionChange` callback should be called in the `handleSelectedChange` function.

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const handleSelectedChange = useCallback(
+    (changedOption: Pick<DropDownOption, 'selected' | 'value' | 'label'>) => {
+      setFilterOptions(prevFilterOptions =>
+        prevFilterOptions.map(option =>
+          option.value === changedOption.value
+            ? { ...option, selected: changedOption.selected }
+            : option
+        )
+      )
+    },
```
fredericvilcot, line 68: You only need here to call the `onSelectionChange` callback.

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,157 @@
...
+        trigger={
+          <div className="okp4-dataverse-portal-dropdown-header">
+            {selectedOptions.length ? (
+              <div className="okp4-dataverse-portal-dropdown-header-selection">
+                {selectedOptions.slice(0, DROPDOWN_TAGS_DISPLAY_LIMIT).map(({ value, label }) => (
+                  <Tag
+                    classes={{ main: 'okp4-dataverse-portal-dropdown-header-selection-item' }}
+                    key={value}
+                    // eslint-disable-next-line react/jsx-no-bind
+                    onDelete={(): void => handleSelectedChange({ selected: false, value, label })}
```
fredericvilcot, line 137: Please, do not use lambda functions but use curryfication here

src/ui/config.ts
```diff
@@ -0,0 +1,2 @@
+export const DROPDOWN_MAX_SEARCH_RESULTS = 10
+export const DROPDOWN_TAGS_DISPLAY_LIMIT = 2
```
fredericvilcot, line 2: These constants are not environment variables, but are declared at build time.
We want them to be replaced at runtime.

fredericvilcot (changes requested):

src/ui/config.ts
ccamel: Yes. As mentionned by @fredericvilcot, consider these values as environment variables injected as runtime. See `APP_ENV` declared in [src/env.d.ts](https://github.com/okp4/dataverse-portal/blob/732902ca92edb2a1437fcc60fd3ea90aae972248/src/env.d.ts).

src/ui/view/share/data/dropDown/dropDown.tsx
oussama-aitbaziz-dev: That what I was doing in the dropped commit I believe that we've discussed that the current implementation should be kept

src/ui/component/checkbox/checkbox.tsx
```diff
@@ -6,22 +6,22 @@ import { Icon } from '@/ui/component/icon/icon'
 import './checkbox.scss'
 
 export type CheckboxOption = {
-  checked: boolean
   value: string
+  checked?: boolean
   id?: string
-  name?: string
+  label?: string
```
0xLodz, line 12: Input type checkbox has a `name` attribute but no `label` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox).
Checkbox component renders only the checkbox input, and DynamicCheckbox renders the `Checkbox` component with an highlighted label.
That's why I recommend to keep the name attribute for both Checkbox and DynamicCheckbox components.

It might be good to rename `DynamicCheckbox` component in a dedicated PR to avoid confusion.

src/ui/view/dataverse/component/dynamicCheckbox/dynamicCheckbox.tsx
```diff
@@ -3,20 +3,26 @@ import { useCallback } from 'react'
...
-import type { SelectionOption } from '@/ui/view/dataverse/component/filters/selectionFilter/selectionFilter'
 import './dynamicCheckbox.scss'
 
+type SelectionOption = {
+  selected: boolean
+  value: string
+  label?: string
+  disabled: boolean
+}
+
```
0xLodz, line 14: I think this type shouldn't be there but moved in another component.

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,137 @@
...
+type DropDownOption = {
+  value: string
+  disabled?: boolean
+  selected?: boolean
+}
+
+type SelectionItemType = 'checkbox'
+
+type DropDownProps = {
+  value: DropDownOption[]
```
0xLodz, line 20: In this context, `value` must be an array of strings. `selected` option state can be managed internally :)

src/ui/view/dataverse/component/dynamicCheckbox/dynamicCheckbox.tsx
```diff
@@ -3,20 +3,26 @@ import { useCallback } from 'react'
...
 type DynamicCheckboxProps = {
-  name: string
+  label?: string
   value: string
-  checked: boolean
+  checked?: boolean
```
0xLodz, line 18: Why these props are now optional?

fredericvilcot (commented):
We are almost done @oussama-aitbaziz-dev , thx!
I just wrote down some minor remarks, it's up to you 😌

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,128 @@
...
+    (searchTerm: string) => {
+      setSearchTerm(searchTerm)
+    },
+    [setSearchTerm]
+  )
+
+  const handleSelectedChange = useCallback(
+    (v: string) => () => {
+      if (value.includes(v)) {
+        return onChange(value.filter(val => val !== v))
```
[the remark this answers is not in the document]
oussama-aitbaziz-dev: Its not useless actually its an early return to avoid executing the add to value part of code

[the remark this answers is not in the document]
fredericvilcot: I see the point 👍
But if you want to respect the fact that the function returns void, I will modify a bit: 
```suggestion
        onChange(value.filter(val => val !== v))
        return
```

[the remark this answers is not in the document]
oussama-aitbaziz-dev: Done <:)

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,125 @@
...
+                    classes={{ main: 'okp4-dataverse-portal-dropdown-header-selection-item' }}
+                    key={v}
+                    onDelete={handleSelectedChange(v)}
+                    tagName={v}
+                  />
+                ))}
+                {value.length > APP_ENV.dropDownTagsDisplayLimit && (
+                  <Tag
+                    classes={{ main: 'okp4-dataverse-portal-dropdown-header-selection-item' }}
+                    tagName={`+${String(value.length)}`}
```
0xLodz, line 112: You need to subtract default display limit. Also, string casting seems unnecessary.
```suggestion
                    tagName={`+${value.length - dropDownTagsDisplayLimit}`}
```

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,125 @@
...
+                <NoResultFound
+                  className="okp4-dataverse-portal-dropdown-no-results-wrapper"
+                  iconName="large-magnifier-with-cross"
+                />
+              )}
+            </div>
+          </div>
+        }
+        iconName="chevron-sharp"
+        open
```
0xLodz, line 96: It seems closed by default.
```suggestion
```

0xLodz (changes requested):

src/ui/view/share/data/dropDown/dropDown.tsx
oussama-aitbaziz-dev: That's what I taught too but when I read the spec, "... a tag containing the total number of selected options appears." I get that it should be the total of selected items

0xLodz: It seems counterintuitive with the leading `+`. We probably need a confirmation from the product team.

oussama-aitbaziz-dev: @0xLodz I've adressed this issue with them, they agree that we both taught is the right implementation

public/assets/env.js
```diff
@@ -73,5 +73,6 @@ const APP_ENV = {
       rest: 'https://api.devnet.staging.okp4.network:443/rpc',
       ...sharedChainConf
     }
-  ]
+  ],
+  dropDown: { dropDownMaxSearchResults: 10, dropDownTagsDisplayLimit: 4 }
```
ccamel, line 77: Although current property names are reasonably clear imho, it is generally desirable to use more descriptive names that correctly reflect their purpose. I would suggest:

```suggestion
  dropDown: { maxDisplayedSearchResults: 10, maxDisplayedTags: 4 }
```

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,157 @@
...
+  onChange: (value: string[]) => void
+  maxSearchResults?: number
+}
+
+const {
+  dropDown: { dropDownMaxSearchResults, dropDownTagsDisplayLimit }
+} = APP_ENV
+
+type DropDownInputProps = Pick<DropDownProps, 'value' | 'placeholder'> & {
+  handleChange: (v: string) => () => void
```
ccamel, line 28: If `handleChange` is intended to be a callback function, it is recommended to make the following adjustments:

- Rename it to `onChange` to align with common naming conventions for callback functions.
- Update its definition to return `void` directly, instead of returning another function `() => void`.

ccamel (changes requested):

src/ui/view/share/data/dropDown/dropDown.tsx
oussama-aitbaziz-dev: The parent component `DropDown` is already getting `onChange`, if I understand correctly your suggestion, this will me break [this lint rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)

ccamel: So, you're telling me that you're twisting the return type to allow the call to be made in the props while respecting `jsx-no-bind rule`? Under no circumstances. Component properties are *public* interfaces that convey meaning and suggest behavior. Specifying a property as a callback that returns a function is nonsensical. The binding should be done in the core of the component, using the most appropriate approach (e.g. `useCallback` or whatever).

oussama-aitbaziz-dev: I've had previous comments about this practice, your suggestion looks more convenient and follows the pattern I've always followed ☺️

ccamel: Ah, I see. Good news, thanks so much for your consideration.

src/ui/component/tag/tag.tsx
```diff
@@ -1,23 +1,38 @@
...
 import { Icon } from '@/ui/component/icon/icon'
 import './tag.scss'
 import { useTranslation } from 'react-i18next'
+import classNames from 'classnames'
 
 type TagProps = {
   tagName: string
-  onDelete?: () => void
+  classes?: { main?: string; name?: string; icon?: string }
+  onDelete?: (v: string) => void
```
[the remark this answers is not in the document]
fredericvilcot: @oussama-aitbaziz-dev 
After more review, why this parameter has been added?
IMHO, it's not necessary.
You can see this implementation without parameter: https://github.com/okp4/dataverse-portal/blob/919d1b3d841de9669962585eeeb31955349fe67e/src/ui/view/tagsField/tagsField.tsx#L39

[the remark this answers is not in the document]
oussama-aitbaziz-dev: @ccamel Suggested this above, I kinda agree with him that's the pattern I always found in react projects, every component should its own handlers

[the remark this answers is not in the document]
fredericvilcot: @oussama-aitbaziz-dev 
@ccamel suggested to remove the currification, not to add a parameter.
Here, you gave back the `tag` value to the parent into the callback parameter that already knows the `value` (because it's the parent that injects the tag value as props)
So, I will remove this useless `v` parameter.

src/ui/view/share/data/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,173 @@
...
+  onChange: (value: string[]) => void
+  maxSearchResults?: number
+}
+
+const {
+  dropDown: { maxDisplayedSearchResults, maxDisplayedTags }
+} = APP_ENV
+
+type DropDownInputProps = Pick<DropDownProps, 'value' | 'placeholder'> & {
+  onChange: (v: string) => void
```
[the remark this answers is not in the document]
fredericvilcot: Thanks, it's better because they have nothing in common.

PR #388, opened 2023-07-12
[FEAT][APPSTORE] Handle errors from ShareData slice's commands

fredericvilcot:
👉 This PR brings some improvements to the previous `ShareFormData` slice.

🧠 The main goal is to handle the possible errors returned by the `initForm` and `setFormItemValue` commands. These errors are accessible from the controller itself and can be logged thanks to the shared `show` function.

🎅 There are also some major improvements, regarding the slice itself which becomes an isolated object entry in the current `appStore`. It also has been renamed to `ShareData` slice that is more straightforward and that will allow to add other properties inside, related to transients states.

🧪 The whole slice has been put under test, and this without the usage of the `react-testing-library` package.


lolottetheclash (approved):
great !!!🚀

src/shared/error.ts
```diff
@@ -0,0 +1,40 @@
+import type { Show } from 'fp-ts/lib/Show'
+
+export const ResourceConflictError = (resourceIds: string[]) =>
+  ({
+    _tag: 'resource-conflict',
+    resourceIds
+  } as const)
+
+/**
+ *  Error when an id of a resource to be stored already exists in memory.
```
ccamel, line 10:
```suggestion
 *  Error when an id of a resource to be stored already exists in the domain.
```

(memory is an implementation detail)

fredericvilcot: Much better 👍

src/ui/store/slice/shareData/test/shareData.slice.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+            const result = store.getState().shareData.setFormItemValue(formItemPayload)()
+            expect(result).toBeEither()
+
+            if (resourceNotFoundError) {
+              expect(result).toBeLeft()
+              expect(result).toEqualLeft(resourceNotFoundError)
+
+              const message = pipe(
+                result,
+                E.getOrElseW(e => ShowFileError.show(e))
```
0xLodz, line 178:
```suggestion
                E.getOrElseW(ShowFileError.show)
```

src/ui/store/slice/shareData/test/shareData.slice.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+            const result = store.getState().shareData.initForm(initialForm)()
+            expect(result).toBeEither()
+
+            if (resourceAlreadyExistsError) {
+              expect(result).toBeLeft()
+              expect(result).toEqualLeft(resourceAlreadyExistsError)
+
+              const message = pipe(
+                result,
+                E.getOrElseW(e => ShowFileError.show(e))
```
0xLodz, line 155:
```suggestion
                E.getOrElseW(ShowFileError.show)
```

0xLodz (approved):
LGTM! 🚀

src/ui/store/slice/shareData/test/shareData.slice.spec.ts
fredericvilcot: Of course! Thx! 😊

src/ui/store/slice/shareData/test/shareData.slice.spec.ts
fredericvilcot: DOne!

ccamel (approved):

PR #389, opened 2023-07-13
[REFACTOR][APPSTORE] Create store from vanilla Zustand instead of React hook

fredericvilcot:
👉 This PR aligns the use of `zustand` between all stores by creating a factory for the `appStore` (from `zustand/vanilla`) instead of directly instantiating a react hook.

🎅 This allows in particular to be able to test the store agnostically to react and thus without having to instantiate the hook with the test library `react-testing-library`.

😱 The test suite for `shareDataForm` has been temporarily removed, because it needs first this PR to be merged.
Will be reimplemented in #388


0xLodz (approved):
LGTM! Using a vanilla store with a factory function appears to be a very convenient approach for testing. Good work!

Issue #247, opened 2023-05-19
[Share Data] Fill in the metadata - General display


ChaimaaZaoui:
As a user, I want to be able to fill in the metadata of the dataset so that it's properly referenced in the dataverse once uploaded.

### Scenario 1: Dataset metadata
**Given that** I am in the step of "Fill in the metadata",
**And** I am using a _desktop_ device,
**Then** the page is displayed as follows

* **Text**  I SHARE A DATASET
* **Figure** Stepper 
* **Text** Fill in the metadata
* **Text** Please complete the dataset metadata for reference purposes. It is highly discouraged to include personal data in the metadata due to its immutability once stored in the blockchain, which does not comply with GDPR regulations.
* **Fields** see the table below
* Button `< Previous step` always active
* Button `Next step >` active when the mandatory fields are filled

| Field title | Field type |Order | Description | Mandatory |
|----|----|----|----|----|
| Dataset title*  | Text | NA | Title with which the dataset will be referenced in the dataverse | Yes |
| Publisher  | Text | NA | entity primarily responsible for making the resource available | No |
| Creator | Text |NA | Dataset creator's name | No |
| Description | Text | NA | Description of what's contained in the dataset, potential uses, data source... | No |
| Format | Drop-down multiselect list | Alphabetical order | list of media types available in the ontology | No |
| License | Drop-down multiselect list | Alphabetical order | referencing the (major) Open Source licenses. list of licenses available in the ontology | No |
| Topic | Drop-down multiselect list | Alphabetical order | Different common topics, allowing resources to be organised and classified according to their nature or purpose. list of Topics available in the ontology | No |
| Geographical coverage | Drop-down multiselect list | Alphabetical order | origin of data in the dataset, list of areas available in the ontology | No |
| Temporal coverage | Date picker | NA | Time range for data application | No |
| Tags | text | NA | Dataset theme | No |
| Pricing | Numerical field  | NA | The minimum fee applicable for the use of the dataset, the field supports only positive real numbers | No |

_the behaviour of the drop down multiselect list is available in the design system_

### Scenario 2 : Format, License, topic, geographical coverage dropdown behaviour
**Given that** I am in the step of "Fill in the metadata",
**And** I want to fill in Format, License, topic, geographical coverage
**when** open a dropdown
**then** there's a search and a multiselect for the available items

**When** I scroll down the list
**then** there is an infinite scroll and loader 
Fetch 20 by 20

**When** an item is selected
[...]

PR #390, opened 2023-07-13
Feat/share-data: create tags field component

0xLodz:
### Behavior
This PR add the TagsField component, allowing the user to add and remove tags. This component is needed for the [share data step 3](https://github.com/okp4/dataverse-portal/issues/247).

### Screenshot
image


src/ui/view/share/data/i18n/share_data_de.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type-press-enter": "Tippen Sie und drücken Sie die <span>Eingabetaste</span>"
```
ccamel, line 2: I don't really like having HTML tags in translations. It's something we can discuss, but the idea of translations is to have something purely textual, even if the [library we use](https://react.i18next.com/latest/trans-component) allows for tag interpolation.

0xLodz: We probably can split this translation in two separated translation, even though the first part of the translation left alone doesn't have that much sense. Using tags for translation is a common practice used in several libraries to isolate a part of it's content for styling. When used regularly (e.g. with `t` from `useTranslation`), it won't create any tag.

src/ui/view/share/data/tagsField/tagsField.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const handleChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      setValue(event.target.value)
+    },
+    [setValue]
+  )
+
+  const handleKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLInputElement>) => {
+      if (event.key === 'Enter') {
```
oussama-aitbaziz-dev, line 27: This might not work in some older browsers we also need to handle `event.keyCode === 13`

fredericvilcot (approved):
LGTM!

src/ui/view/share/data/tagsField/tagsField.tsx
0xLodz: Which browser are you thinking of? I think the support is good for the required browsers. https://caniuse.com/?search=event.key

oussama-aitbaziz-dev: Im used to handling key codes and key to make sure everything works everywhere :)

src/ui/view/share/data/tagsField/tagsField.scss
```diff
@@ -0,0 +1,62 @@
...
+
+    .okp4-dataverse-portal-tag-main {
+      min-height: $tag-height;
+    }
+
+    .okp4-dataverse-portal-tags-field {
+      position: relative;
+      height: $tag-height;
+
+      &-placeholder {
```
lolottetheclash, line 27: This way to handle classnames works fine, but to stay in line with the way we manage classnames in general, can you keep our original way of doing things?

lolottetheclash (changes requested):
Nice job! It's looking great!
Some minor changes, and one to keep in mind for v2:
currently, the way to manage tags is based on the value, which is a string.
The problem is that we can add the same tag twice, and when we want to delete one of them, both disappear.
It will be interesting (for V2) to check with the product team whether it's allowed to add the same tag twice, and if not, how to signal to the user that this isn't allowed. 
If it is allowed, we'll need to think about modifying the tag data format to add an id, for example.

lolottetheclash (changes requested):

src/ui/view/share/data/tagsField/tagsField.tsx
```diff
@@ -0,0 +1,71 @@
...
+        const tag = event.currentTarget.value.trim()
```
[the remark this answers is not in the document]
0xLodz: I agree that it improves code readability. The code was updated with a renamed value state `inputValue`.

src/ui/view/share/data/i18n/share_data_de.json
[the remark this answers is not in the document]
0xLodz: Oops. Thanks!

src/ui/view/share/data/tagsField/tagsField.tsx
```diff
@@ -0,0 +1,71 @@
...
+      ))}
+      <div className="okp4-dataverse-portal-tags-field">
+        <span
+          className={classNames('okp4-dataverse-portal-tags-field-placeholder', { hidden: value })}
+        >
+          <Trans
+            components={{
+              span: <span className="okp4-dataverse-portal-tags-field-placeholder-action" />
+            }}
+            i18nKey="shareData:type-press-enter"
```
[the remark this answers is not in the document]
0xLodz: Good catch. Fixed in [7659241](https://github.com/okp4/dataverse-portal/pull/390/commits/76592413368f55b72abfc9c6d6e175648d36a68b)!

0xLodz:
> Nice job! It's looking great! Some minor changes, and one to keep in mind for v2: currently, the way to manage tags is based on the value, which is a string. The problem is that we can add the same tag twice, and when we want to delete one of them, both disappear. It will be interesting (for V2) to check with the product team whether it's allowed to add the same tag twice, and if not, how to signal to the user that this isn't allowed. If it is allowed, we'll need to think about modifying the tag data format to add an id, for example.

Thanks for your feedback, we will have a discussion with the product team for your suggested improvements.
I think similar tags that disappear can be handled by the parent.
But yes, adding the same tag twice is the responsibility of this component. 👍

src/ui/view/share/data/i18n/share_data_de.json
0xLodz: We finally split the translation and removed the tag.

lolottetheclash (approved):
Great job, thanks !

PR #401, opened 2023-07-17
Feat/Field Component V2

0xLodz:
## Field component V2

This PR aims to improve the generic Field component in order to fit with Pricing field requirements for [step 3](https://github.com/okp4/dataverse-portal/issues/247) and step 4 of share data. It adds the styles for fields with placeholders

- [x] Add readonly state
- [x] Make label optional
- [x] Add placeholder
- [x] Adjust label and placeholder styles

🚨🚨 This PR doesn't implement the numerical input rules, those are saved for later.

## Screenshot

Screenshot 2023-07-19 at 10 35 38

### DIFF
--- a/src/ui/component/field/field.tsx
+++ b/src/ui/component/field/field.tsx
@@ -5,7 +5,7 @@ import './field.scss'
 
 type FieldProps = {
   id: string
-  label: string
+  label?: string
   onChange: (value: string) => void
   value?: string
   type?: 'text' | 'number'
@@ -95,12 +95,16 @@ export const Field: FC<FieldProps> = ({
           {rightElement}
         </div>
       )}
+
+      {label && (
       <label
         className={classNames(textFieldClassNames, 'okp4-dataverse-portal-field-label')}
         htmlFor={id}
       >
         {required ? label + '*' : label}
       </label>
+      )}
+
       {error && (
         <p className="okp4-dataverse-portal-field-error">
           <Icon name="info-light" />
--- a/src/ui/component/field/field.tsx
+++ b/src/ui/component/field/field.tsx
@@ -12,6 +12,7 @@ type FieldProps = {
   error?: string
   required?: boolean
   disabled?: boolean
+  readonly?: boolean
   multiline?: boolean
   resizable?: boolean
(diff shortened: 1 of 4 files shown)


AnjaMis (approved):

src/ui/component/field/field.tsx
```diff
@@ -1,21 +1,23 @@
...
   error?: string
+  placeholder?: string
   required?: boolean
   disabled?: boolean
+  readonly?: boolean
   multiline?: boolean
   resizable?: boolean
   leftElement?: JSX.Element
-  rightElement?: JSX.Element
+  rightElement?: ReactNode
```
fredericvilcot, line 20: Just for my comprehension, why `rightElement` becomes a `ReactNode`?
Be careful with `ReactNode` because it accepts everything: 

`type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;`

By the way, why `leftElement` remains a `JSX.Element` ?

fredericvilcot (commented):
Great, just a minor interrogation 😌

src/ui/component/field/field.tsx
AnjaMis: I'm not sure about the all possible `leftElement` options, so didn't want to change that yet. For the `rightElement` we know already that there is a case where we need to add `string`, such as 'KNOW' in the step 3 and 4. We tried wrapping that text in <span> tag initially, but it was not matching the style we needed so we thought of replacing JSX.Element to ReactNode. Would you say this is not ideal solution, should I try something else?

0xLodz: Is `JSX.Element | string` a more suitable alternative for limiting the `rightElement` type to what we need?

fredericvilcot: By the way, if you need a string, you can keep a single JSX.Element by doing this instead:

```tsx
rightELement={<span>KNOW</span>}
```

AnjaMis: Fixed now!

fredericvilcot (approved):
LGTM 😎

PR #403, opened 2023-07-19
feat(metadataFilling): add required fields validation

lolottetheclash:
This PR adds validation to step 3 of the share data: only the dataset title field is mandatory.

Validation remains in the UI for the time being, pending implementation at domain level with the appropriate business rules. 

🙏 For testing, just add an extra step after step 3 in order to see disabled button goes enabled after filling title only mandatory field.

### DIFF
--- a/src/ui/page/share/dataset/shareDataset.tsx
+++ b/src/ui/page/share/dataset/shareDataset.tsx
@@ -5,17 +5,45 @@ import { DataSelection } from './steps/dataSelection/dataSelection'
 import { MetadataFilling } from './steps/metadataFilling/metadataFilling'
 import { Stepper } from '@/ui/component/stepper/stepper'
 import type { StepElement } from '@/ui/component/stepper/stepper'
+import { useAppStore } from '@/ui/store'
+import type { DatePickerValue, FormItem } from '@/ui/store/slice/shareData/shareData.slice'
 import '../i18n/index'
 import './shareDataset.scss'
 
 export const ShareDataset: FC = () => {
   const { t } = useTranslation('share')
-
+  const { form } = useAppStore(state => ({
+    form: state.shareData.form
+  }))
   const storageServiceSelection: StepElement = { id: 'step1', content: <ServiceStorageSelection /> }
   const dataSelection: StepElement = { id: 'step2', content: <DataSelection /> }
   const metadataFilling: StepElement = { id: 'step3', content: <MetadataFilling /> }
 
   const steps = [storageServiceSelection, dataSelection, metadataFilling]
+
+  const isFormItemFilled = (item: FormItem): boolean => {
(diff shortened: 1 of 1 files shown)


src/ui/page/share/dataset/shareDataset.tsx
```diff
@@ -5,15 +5,48 @@ import { DataSelection } from './steps/dataSelection/dataSelection'
...
+    case 'numeric': {
+      const numericField = item.value as unknown as number
+      return numericField === 0 || !!numericField
+    }
+    case 'date': {
+      const datePickerField = item.value as unknown as DatePickerValue
+      return !!datePickerField.from || !!datePickerField.to
+    }
+  }
+}
```
ccamel, line 30: 😵

 I'm a little too late, but the problem here lies in the following type definitions:

```ts
type FormItemType = 'text' | 'numeric' | 'date' | 'select'
type FormItemValue = string | string[] | number | DatePickerValue

type FormItem = Readonly<{
  id: FormItemId
  type: FormItemType
  label: string
  value: FormItemValue
  required?: boolean
}>
```

This definitely illustrates the problem with how we're defining the `FormItem` type, which does not embrace an [Algebraic Data Type](https://en.wikipedia.org/wiki/Algebraic_data_type) approach. Firstly, we are allowing representations that should not be possible, which can lead to invalid states (rule:  [Making illegal states unrepresentable](https://fsharpforfunandprofit.com/posts/designing-with-types-making-illegal-states-unrepresentable/)). Secondly, we're not leveraging discriminated unions, which would allow us to safely manipulate the values in a type-safe manner.

0xLodz: Good point. We could update FormItem time as follow:
```ts
type FormItem = {
  type: 'text';
  value: string;
} | {
  type: 'select';
  value: string[];
} | {
  type: 'numeric';
  value: number;
} | {
  type: 'date';
  value: DatePickerValue;
} | ...
```

Thus simplify the `isFormItemFilled ` function:
```ts
const isFormItemFilled = (item: FormItem): boolean => {
  switch (item.type) {
    case 'text': {
      return !!item.value
    }
    case 'select': {
      return item.value.length > 0
    }
    case 'numeric': {
      return item.value === 0 || !!item.value
    }
    case 'date': {
      return !!item.value.from || !!item.value.to
    }
  }
}
```

ccamel: Oh my! Exactly what should be done @0xLodz :relieved:  What I am reading is a delight to my eyes and fills me with an ineffable joy.

lolottetheclash: Indeed it's much more elegant and safer, well seen @ccamel  and well coded @0xLodz !

Issue #248, opened 2023-05-19
[Share Data] All in summary - General display


ChaimaaZaoui:
As a user, I want to be able to read the summary of dataset sharing process so that I can edit any information before proceeding to the payment 

### Scenario 1: Metadata and file overview

**Given that** I am in the step of "All in summary",
**And** I am using a *desktop* device,

**Then** I can have an overview of the files I selected in my dataset
**And** the the metadata I filled for the dataset 
**And** the page is displayed as follows

- **Text**  _I SHARE A DATASET_
- **Figure** Stepper 
- **Text** _All in summary_
- **Metadata section ** + scroll bar
-- Dataset title, Publisher, Creator, Description, format, license, topic, geographical coverage, temporal coverage, tags
- **List** of files, with a scroll bar if the list doesn't fit within the designated area
- **Price field** + **Text** _Fee set for the use of this dataset_
- **Checkbox** + **Text** _I certify that I do not share any personal metadata._ 
- Button `< Previous step` always active
- Button `Next step >` active when the checkbox is ticked

### Scenario 2 : Internationalisation 
ENG
> All in summary

> Dataset title
> Publisher
> Creator
> Description
> Format
> License 
> Topic
> Geographical coverage
> Temporal coverage
> Tags 

> Fees set for the use of the dataset

> I certify that I do not share any personal metadata

FR
> Récapitulatif

> Titre du jeu de données
> Éditeur
> Créateur
> Description
> Format
> Licence 
> Thème
> Couverture géographique
> Couverture temporelle
> Tags 

> Redevances liées à l'utilisation des données

> Je certifie que je ne partage aucune métadonnée personnelle.

DE
> Zusammenfassung 

> Titel des Datensatzes
> Herausgeber
> Ersteller
> Beschreibung
> Format
> Lizenz 
> Thema
> Geografische Abdeckung
> Zeitliche Abdeckung
> Tags 

> Gebühren für die Nutzung der Daten

> Ich bestätige, dass ich keine persönlichen Metadaten weitergebe.

#### Desktop
![image.png](https://images.zenhubusercontent.com/642d29f6f41891784a960ff2/967a3d3a-6214-45b0-9107-acd00a99adda)

### Environment

- staging URL: <https://app.staging.okp4.network/>

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206526720) by [Unito](https://www.unito.io)

ChaimaaZaoui:
[...]

PR #404, opened 2023-07-19
Feat/add share dataset summary

AnjaMis:
This PR adds a 4th step in the data sharing.

Issue: #248 

## Screenshots
<details>

<summary> Large screen </summary>

Screenshot 2023-07-20 at 14 28 29

</details>

<details>
<summary> Medium screen </summary>

Screenshot 2023-07-20 at 14 31 36

</details>

### Dark theme with tags
image

## Notes

- the `onChange` property in the Field component was made optional because of the added readonly fields that cannot be changed


src/ui/component/field/field.tsx
```diff
@@ -6,7 +6,7 @@ import './field.scss'
 type FieldProps = {
   id: string
   label?: string
-  onChange: (value: string) => void
+  onChange?: (value: string) => void
```
oussama-aitbaziz-dev, line 9: I think `onChange` should never be optional in a input component

src/ui/page/share/dataset/steps/summary/summary.scss
```diff
@@ -0,0 +1,101 @@
...
+    padding: 34px 10px 0 0;
+    box-sizing: border-box;
+    overflow-y: auto;
+
+    @include use-breakpoints(('desktop')) {
+      @include grid-container($colGutter: 34px, $rowGutter: 26px);
+    }
+
+    h2 {
+      @include grid-item($colStart: 1, $colEnd: -1, $rowStart: auto, $rowEnd: auto);
```
oussama-aitbaziz-dev, line 19:
```suggestion
      @include grid-item($colStart: 1, $colEnd: -1);
```
The mixin has been updated to replace undefined arguments as `auto`

src/ui/page/share/dataset/steps/summary/summary.scss
```diff
@@ -0,0 +1,101 @@
...
+        padding-right: 14px;
+        max-height: 430px;
+
+        @include use-breakpoints(('desktop')) {
+          max-height: 360px;
+        }
+      }
+    }
+
+    .okp4-dataverse-portal-share-dataset-summary-certify {
```
oussama-aitbaziz-dev, line 90: This element should have more vertical margin between the summary and step nav bar

src/ui/page/share/dataset/steps/summary/summary.scss
```diff
@@ -0,0 +1,101 @@
+@import '@/ui/style/fonts.scss';
+@import '@/ui/style/theme.scss';
+@import '@/ui/style/mixins';
+
+.okp4-dataverse-portal-share-dataset-summary-container {
+  @include with-theme {
+    @include grid-container($colGutter: 48px, $rowGutter: 22px);
+    grid-template-rows: auto 1fr 1fr;
```
oussama-aitbaziz-dev, line 8: I think this is not needed since grid childen have `grid-row-start: auto`

src/ui/view/share/knowFee/knowFee.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div className={classNames('okp4-dataverse-portal-share-data-know-fee-main', theme)}>
+      <p>{label}</p>
+      <div className="okp4-dataverse-portal-share-data-know-fee-field">
+        <Field
+          id="know-fee"
+          placeholder="0"
+          readonly={readonly}
+          rightElement={<span>KNOW</span>}
+          value={String(fee?.value)}
```
0xLodz, line 27: If `fee` is not defined, the input displays `"undefined"`.

src/ui/view/share/knowFee/knowFee.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export const KnowFee: FC<KnowFeeProps> = ({ readonly = false, label }) => {
+  const theme = useAppStore(store => store.theme)
+
+  const formItems = useAppStore(state => state.shareData.form)
+  const fee = formItems.find(item => item.label === 'fee')
+
+  return (
+    <div className={classNames('okp4-dataverse-portal-share-data-know-fee-main', theme)}>
+      <p>{label}</p>
```
0xLodz, line 20: Using the `<label>` tag might me more appropriate here.

src/ui/page/share/dataset/steps/summary/summary.scss
0xLodz, line 90:
```suggestion
    .okp4-dataverse-portal-share-dataset-summary-certify {
       cursor: pointer;
```

0xLodz (changes requested):
Great work! Provided you some minor request changes ☺️

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <div className="okp4-dataverse-portal-share-dataset-summary-container">
+      <h2>{t('share:share.dataset.summary')}</h2>
+      <div className="okp4-dataverse-portal-share-dataset-summary-left">
+        <div className="okp4-dataverse-portal-share-dataset-summary-text-container">
+          {filteredformItems.map(({ label, value }, index) => (
+            <div className="okp4-dataverse-portal-share-dataset-summary-text" key={index}>
+              <h3>{label.replace(/([A-Z])/g, ' $1')}</h3>
+              <span>{value.toString()}</span>
```
oussama-aitbaziz-dev, line 44: `<Tag />` should be used here I guess?

Capture d’écran 2023-07-20 à 15 18 15

oussama-aitbaziz-dev: `value.toString()` is a potential bug since `value` answers to this type `type FormItemValue = string | string[] | number | DatePickerValue`

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const Summary: FC = () => {
+  const { t } = useTranslation('share')
+  const { files } = useFileStore(state => ({
+    files: state.filesDescriptor
+  }))
+
+  const certifyText = t('share:share.dataset.summaryCertify')
+
+  const [isChecked, setChecked] = useState<boolean>(false)
```
oussama-aitbaziz-dev, line 19:
```suggestion
  const [isChecked, setChecked] = useState(false)
```

src/ui/view/share/knowFee/knowFee.tsx
```diff
@@ -0,0 +1,32 @@
...
+type KnowFeeProps = {
+  readonly?: boolean
+  label: string
+}
+
+export const KnowFee: FC<KnowFeeProps> = ({ readonly = false, label }) => {
+  const theme = useAppStore(store => store.theme)
+
+  const formItems = useAppStore(state => state.shareData.form)
+  const fee = formItems.find(item => item.label === 'fee')
```
oussama-aitbaziz-dev, line 16:
```suggestion
  const fee = useAppStore(( shareData: { form }) => state.shareData.form).find(item => item.label === 'fee')
```
Its only a suggestion, feel free to self resolve if you dont agree

src/ui/page/share/dataset/steps/summary/summary.tsx
AnjaMis: Hmm I wasn't sure how else to deal with it because it was throwing me an error that `DatePickerValue` is not a `ReactNode`. Any suggestions?

src/ui/component/field/field.tsx
AnjaMis: What about in cases when the field is `readonly`? That's why I changed it... :/

lolottetheclash: optional is good :)

src/ui/view/share/knowFee/knowFee.tsx
AnjaMis: Good catch, thank you!

src/ui/view/share/knowFee/knowFee.tsx
AnjaMis: Hmm that kinda messes up the style, maybe we keep it as a `<p>` ?

src/ui/page/share/dataset/steps/summary/summary.scss
AnjaMis: Nice catch, fixed!

src/ui/view/share/knowFee/knowFee.tsx
AnjaMis: Ah I tried but it throws a lot of errors so I prefer to keep it simple if you don't mind :)

fredericvilcot: This one can't work because `useAppStore` expects a specific selector function that returns the memoized slice property.
It the way `zustand` works in its core.

src/ui/page/share/dataset/steps/summary/summary.tsx
0xLodz: @oussama-aitbaziz-dev good catch! Tags has been added in [606fb67](https://github.com/okp4/dataverse-portal/pull/404/commits/606fb670dfb867917798584fcc25b8376ef85ee6).

0xLodz (approved):

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -0,0 +1,99 @@
...
+  return (
+    <div className="okp4-dataverse-portal-share-dataset-summary-container">
+      <h2>{t('share:share.dataset.summary')}</h2>
+      <div className="okp4-dataverse-portal-share-dataset-summary-left">
+        <div className="okp4-dataverse-portal-share-dataset-summary-item-container">
+          {filteredformItems.map(({ label, value }, index) => {
+            return (
+              <div className="okp4-dataverse-portal-share-dataset-summary-item" key={index}>
+                <h3 className="okp4-dataverse-portal-share-dataset-summary-item-title">
+                  {label.replace(/([A-Z])/g, ' $1')}
```
ccamel, line 50: Why this trick?

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -0,0 +1,99 @@
...
+                </h3>
+                {Array.isArray(value) ? (
+                  <div
+                    className={classNames(
+                      'okp4-dataverse-portal-share-dataset-summary-item-list',
+                      label + '-label'
```
ccamel, line 56:
```suggestion
                      `{label}-label`
```

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    )}
+                  >
+                    {value.map((tag: string, index: number) => (
+                      <Tag key={`${tag}-${index}`} tagName={tag} />
```
ccamel, line 60: Stable and unique identifiers for keys is generally preferred. In this context, `tags` can be considered as value objects that possess uniqueness within the set. So, you can utilize the tag itself as the identifier for the key prop imho. You can add an extra layer of security by removing duplicates from the array of tags if you want.

ccamel, line 60: FWIU, because the type is an array, you're assuming it's a list of tags? That's extremely weak. :disappointed:  I confess I don't clearly understand the logic of mapping here, or the logic adopted, which may be circumstantial. That said, it's probably related to my comment about form types.

ccamel (changes requested):

src/ui/page/share/dataset/steps/summary/summary.tsx
0xLodz: This trick allows camelCase label to be displayed as separated words. Do these labels have to be internationalised?

src/ui/page/share/dataset/steps/summary/summary.tsx
0xLodz: Thanks! Using a Set seems more appropriate here.

src/ui/page/share/dataset/steps/summary/summary.tsx
ccamel: Firstly, I don't understand why `labels` are written in camel case. The generally accepted understanding about labels is to use them as human-readable versions of elements (such as form items in our case). Consequently, a label should not serve as a key or be used for identification purposes. Secondly, yes labels, just like other elements, require internationalization.

So, if we disregard i18n, `labels` should be used in the UI exactly as they are, with no other transformation except for stylistic changes. However, if we want to implement i18n, then the `label` values should be declared in translation files.

0xLodz: Sorry for the dumb question about i18n. We will update it by providing the needed translations.

PR #406, opened 2023-07-19
Feat/add selected service to share data slice

lolottetheclash:
This PR removes the `selectedStorage` in share data 1st step which was stored in a local state, and add it into the `shareData` slice.

### DIFF
--- a/src/ui/store/slice/shareData/shareData.slice.ts
+++ b/src/ui/store/slice/shareData/shareData.slice.ts
@@ -1,19 +1,20 @@
 /* eslint-disable max-lines-per-function */
+import { flow, pipe } from 'fp-ts/lib/function'
+import { contramap as eqContramap } from 'fp-ts/Eq'
+import { ResourceAlreadyExistsError, ResourceNotFoundError } from '@/shared/error'
 import type { StateCreator } from 'zustand'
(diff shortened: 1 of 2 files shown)


ccamel (approved):

src/ui/store/slice/shareData/shareData.slice.ts
```diff
@@ -38,13 +39,17 @@ export type SetFormItemValuePayload = { id: FormItemId; value: FormItemValue }
 export type initFormItem = FormItem
 export type initFormPayload = initFormItem[]
 
+export type storageServiceId = string
+
 export type ShareDataSlice = {
   shareData: {
     form: Form
+    storageService: O.Option<storageServiceId>
```
fredericvilcot, line 47: As we handle an ID, it would have been more accurate to rename it `storageServiceId` and so on for the setter function 😌

fredericvilcot (commented):
Great, just a minor remark about property names.

0xLodz (approved):
Great!

src/ui/store/slice/shareData/shareData.slice.ts
lolottetheclash: yes sure!

PR #407, opened 2023-07-19
Feat/sparql service category filter

fredericvilcot:
👉 This PR  aims to add a new filter regarding the `sparql` query onto the ontology database.

🧠 This new `byServiceCategoryFilter` allows to constraint the results for the share data - step1 (select a service storage) only to the objects in database which `serviceCategory` strictly equals `Storage` from the `vocabulary-category-service`.

🌶️ Technically, this PR adds 2 commands [`setByServiceCategoryFilter`- `resetByServiceCategoryFilter`] which return both `IO<void>` and add the desired filter onto the `sparql` gateway.

### DIFF
--- a/src/domain/dataverse/domain.ts
+++ b/src/domain/dataverse/domain.ts
@@ -98,7 +98,7 @@ export const storeFactory = (
             limit: 20,
             hasNext: false,
             error: O.none,
-            filters: { byType: 'all', byProperty: null, byServiceCategory: O.none },
+            filters: { byType: 'all', byProperty: O.none, byServiceCategory: O.none },
             language: 'en'
           }))
         ),
@@ -112,11 +112,8 @@ export const storeFactory = (
             () => get().data.filters.byProperty,
             IO.map(
               flow(
-                O.fromNullable,
-                O.match(
-                  () => '',
-                  filter => filter.value
-                )
+                O.map(f => f.value),
+                O.getOrElse(() => '')
               )
             )
           ),
@@ -212,8 +209,8 @@ export const storeFactory = (
                   newFilter.value,
                   S.isEmpty,
                   B.match(
-                    () => newFilter,
-                    () => null
+                    () => O.some(newFilter),
+                    () => O.none
                   )
                 )
               }
@@ -237,7 +234,7 @@ export const storeFactory = (
               dataverse: [],
               filters: {
                 ...state.data.filters,
-                byProperty: null
+                byProperty: O.none
               }
             }
           })),
--- a/src/domain/dataverse/domain.ts
+++ b/src/domain/dataverse/domain.ts
@@ -16,6 +16,7 @@ import type { Dataverse } from './entity'
 import type { DataversePort } from './port'
 import type {
   ByPropertyFilterInput,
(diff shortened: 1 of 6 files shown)


0xLodz (approved):
LGTM! 🚀

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -39,6 +41,7 @@ export const sparqlGateway: DataversePort = {
       PREFIX datasetMetadata: <https://ontology.okp4.space/metadata/dataset/>
       PREFIX zoneMetadata: <https://ontology.okp4.space/metadata/zone/>
       PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+      PREFIX svccat: <https://ontology.okp4.space/thesaurus/service-category/>
```
ccamel, line 44: To maintain consistency, I would recommend to standardize the style of prefixes within `RDF` ontologies. In your code, camel case style was used, while here lower case style is being utilized. Moreover, although prefixes can be named arbitrarily, it is generally a good thing to adopt standardized prefixes to ensure clarity and recognition (e.g. `foaf`, `rdfs`, `cc`).

I think we should rename the other prefixes to camel case for all prefixes.

lolottetheclash (approved):
Well done! 🚀

src/infra/dataverse/sparql/sparqlGateway.ts
```diff
@@ -56,6 +59,7 @@ export const sparqlGateway: DataversePort = {
         }
         ${byType === 'all' ? '' : `FILTER ( ${byTypeFilter(byType)} )`}
         ${byProperty?.property ? `FILTER ( ${byPropertyFilter(byProperty)} )` : ''}
+        ${O.isSome(byServiceCategory) ? `FILTER ( ?topic = svccat:${byServiceCategoryFilter})` : ''}
```
ccamel, line 62: It's indeed correct, but I would have seized the opportunity to chain options, with a `O.map` + a `O.getOrElse` that I fill is more functional. However, this decision ultimately comes down to personal preference. :wink:

ccamel (changes requested):
That's fine, I just have a few minor remarks that I'll let you appreciate.

src/infra/dataverse/sparql/sparqlGateway.ts
fredericvilcot: Yes, of course 😌
I'll suggest to follow the `lowercase` convention used on the ontology classes.

src/domain/dataverse/domain.ts
```diff
@@ -97,7 +98,7 @@ export const storeFactory = (
             limit: 20,
             hasNext: false,
             error: O.none,
-            filters: { byType: 'all', byProperty: null },
+            filters: { byType: 'all', byProperty: null, byServiceCategory: O.none },
```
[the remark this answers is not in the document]
fredericvilcot: Yep, you're right.
My first plan was to refactor code in a next PR, but we can put it here as there is no a huge impact on changes.

ccamel (approved):
Great! Thanks :+1:

PR #408, opened 2023-07-20
Feat/Numerical Field

0xLodz:
## Numerical field
When the user is typing, the field must be formatted with narrow no-break space as thousands separators and full stop as decimal separator.

### Usage
```ts
 <NumericalField
    decimalLimit={APP_ENV.chains[0].feeCurrencies[0].coinDecimals}
    decimalSeparator="."
    id="fees"
    // locale="de-DE"
    onChange={handleChange}
    placeholder="0"
    rightElement={<span>{APP_ENV.chains[0].currencies[0].coinDenom}</span>}
    value={value}
  />
```


src/util/field/field.ts
```diff
@@ -0,0 +1,37 @@
+import * as O from 'fp-ts/Option'
+import { pipe } from 'fp-ts/function'
+import { escapeRegExp } from '@/util/util'
+
+/**
+ * Returns a function that checks if a given string s is a valid number field format.
```
ccamel, line 6: Here, I wouldn't call it a "field". The purpose of the function is to check that a given string is a valid number (with a few assumptions), whatever its use case.

src/util/field/field.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param {string} config.thousandSeparator - The character used as the thousand separator.
+ * @param {string} config.decimalSeparator - The character used as the decimal separator.
+ * @param {number} config.decimalLimit - The maximum number of digits that can follow the decimal separator.
+ * @returns {(s: string) => O.Option<string>} - A function that checks if a given string s is a valid number format according to the specified configuration.
+ * Returns an fp-ts Option of the string if it is valid, and fp-ts None otherwise.
+ * @example isValidNumberFieldFormat({ thousandSeparator: '\u202F', decimalSeparator: '.', decimalLimit: 2 })('123.') // some('123.')
+ * @example isValidNumberFieldFormat({ thousandSeparator: '\u202F', decimalSeparator: ',', decimalLimit: 2 })('123.1') // some('123.1')
+ */
+export const isValidNumberFieldFormat =
+  ({
```
ccamel, line 18: It's quite unusual for a (utilitary) function to take an object as argument, and not positional parameters, imho.

src/util/field/field.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param {{ thousandSeparator: string; decimalSeparator: string; decimalLimit: number }} config - The configuration object.
+ * @param {string} config.thousandSeparator - The character used as the thousand separator.
+ * @param {string} config.decimalSeparator - The character used as the decimal separator.
+ * @param {number} config.decimalLimit - The maximum number of digits that can follow the decimal separator.
+ * @returns {(s: string) => O.Option<string>} - A function that checks if a given string s is a valid number format according to the specified configuration.
+ * Returns an fp-ts Option of the string if it is valid, and fp-ts None otherwise.
+ * @example isValidNumberFieldFormat({ thousandSeparator: '\u202F', decimalSeparator: '.', decimalLimit: 2 })('123.') // some('123.')
+ * @example isValidNumberFieldFormat({ thousandSeparator: '\u202F', decimalSeparator: ',', decimalLimit: 2 })('123.1') // some('123.1')
+ */
+export const isValidNumberFieldFormat =
```
ccamel, line 17: Well, I'm in trouble with the naming of the function if this one returns a function. If it's a higher order function, it should be named `numberFormatValidator` for instance.

src/ui/component/field/numericalField.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { pipe, flow } from 'fp-ts/function'
+import type { FieldProps } from '@/ui/component/field/field'
+import { Field } from '@/ui/component/field/field'
+import { isValidNumberFieldFormat } from '@/util/field/field'
+import { isEmptyOrEndsWithDot } from '@/util/util'
+import { formatLocalizedNumber } from '@/util/i18n/i18n'
+
+type NumericalFieldProps = FieldProps & {
+  onChange: (value: string) => void
+}
```
ccamel, line 13: I don't like too much this because it creates a duplicate definition of the `onChange` prop, which could lead to maintenance issues if the `FieldProps` definition changes.

I'd like to suggest this change:

```suggestion
type NumericalFieldProps = FieldProps & Required<Pick<FieldProps, 'onChange'>>;
```

This allows all changes to be inherited from the `onChange` prop in `FieldProps`, ensuring that both types remain in sync, imho.

ccamel (changes requested):

lolottetheclash (changes requested):
Nice work with tests and fp-ts! 💪

I have some remarks, don't hesitate if you want to discuss about it:

about genericity: I think we should'nt hardcode the fact that `,`separator must be transformed into `.`.
Also, not hardcode the use of `coinDecimal` env, neither hardcode the `fr` locale.
I think this component should move towards a more generic component.

Also, when testing, it was impossible for me to enter any number after the decimal; and when I wanted to delete a number (before the decimal point) among the others, it deletes it for me and then the cursor returns to the end...
Which is annoying because if you want to delete two numbers quickly, you don't realize that you're deleting the first (wanted) number and the second deleted number is the last (unwanted) number.

To enter digits after the decimal point, it's impossible for me to enter a `0`, so I think the aim is to avoid ending the user's input with a zero, but if you want to enter 10.505 for example, it's impossible.

src/ui/component/field/numericalField.tsx
0xLodz: Beautiful! ✨ Thanks!

0xLodz:
Given the intricate nature of this feature, particularly the myriad scenarios to take into account when handling the display and editing of formatted input content, we have discussed about the most appropriate alternative to create the Numerical Input component from scratch. Factors under consideration include:

- Adjusting text size in relation to the quantity of thousand separators
- Maintaining cursor position
- Handling selection of multiple characters for deletion or addition of content
- Distinguishing between the usage of Backspace or Delete keys on Windows
- Managing negative numbers, leading zeroes, etc

Given these complexities, we concluded that using a library dedicated to numerical input is the most effective and efficient approach, and probably rebuild your own numerical field in the future according to our needs. Therefore, we have closed this PR in favour of a new implementation. You can find the updated feature implementation in PR #428.

PR #410, opened 2023-07-20
feat(metadataFilling): add tagsField to form

lolottetheclash:
This PR adds the tags field into the fill metadata step 3 form.


fredericvilcot (approved):
🫶

PR #428, opened 2023-07-28
Feat/numerical field

0xLodz:
## Numerical field
When the user is typing, the field is be formatted according to the provided locale (`lang`), separator can be overridden using `thousandSeparator` and `decimalSeparator` props.

In order to comply with product requirement for Fees input, we define the narrow no-break space as thousands separator and full stop as decimal separator. Also, `decimalScale` allow to get the expected number for the KNOW token.

### Usage
```ts
<Field
  allowLeadingZeros={false}
  allowNegative={false}
  decimalScale={APP_ENV.chains[0].feeCurrencies[0].coinDecimals} // 6
  decimalSeparator="."
  id="fees"
  lang="fr-FR" 
  onChange={handleChange}
  placeholder="0"
  rightElement={<span>{APP_ENV.chains[0].currencies[0].coinDenom}</span>} // KNOW
  thousandSeparator=" " // narrow non-breaking spaces
  type="numeric" // 👈 define field type for numeric input
  value={value}
/>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-markdown": "^8.0.7",
+    "react-number-format": "^5.2.2",
     "react-router-dom": "^6.14.1",
     "react-transition-group": "^4.4.5",
     "short-uuid": "^4.2.2"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,6 +17,9 @@ dependencies:
   react-markdown:
     specifier: ^8.0.7
     version: 8.0.7(@types/react@18.2.15)(react@18.2.0)
+  react-number-format:
+    specifier: ^5.2.2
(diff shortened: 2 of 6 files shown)


0xLodz:
Given the intricate nature of this feature, particularly the myriad scenarios to take into account when handling the display and editing of formatted input content, we have discussed about the most appropriate alternative to create the Numerical Input component from scratch. Factors under consideration include:

- Adjusting text size in relation to the quantity of thousand separators
- Maintaining cursor position
- Handling selection of multiple characters for deletion or addition of content
- Distinguishing between the usage of Backspace or Delete keys on Windows
- Managing negative numbers, leading zeroes, etc

Given these complexities, we concluded that using a library dedicated to numerical input is the most effective and efficient approach, and probably rebuild your own numerical field in the future according to our needs. Therefore, we have closed this PR in favour of a new implementation. You can find the updated feature implementation in PR #428.

PR #429, opened 2023-07-28
Feat/vocab domain

fredericvilcot:
👉 This PR brings the first implementation of the `Vocabulary` domain.

🧠 This consists in asking the vocabulary from ontology thanks to a `retrieveVocabularyByType` command which allows to select which vocabulary you want to retrieve (between `license`, `topic`, `media-type` and `area`)

⚠️ It remains a lot of things to implement:

- remove the `abort` mechanism in the `fetch` methods in favor of mutex locks
- add support for a filter by property that will allow searches by characters into the ontology
- add a `retrieveVocabularyById` command that will handle the retrieval of vocabulary elements by their ids (e.g retrieve the elements selected in a multi select)

🎅 In addition, it also adds:

- a complete refactor of error management, by creating some shared error constructors and types
- following these news errors, a refactor of the `loadDataverse` command from `dataverse` domain
- helper functions that allow to easily build some `Reader` dependencies, following the [_functional option pattern_](https://golang.cafe/blog/golang-functional-options-pattern.html)
- tests for `commands` and `helpers`

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206524564) by [Unito](https://www.unito.io)

### Summary by CodeRabbit

- New Feature: Introduced a vocabulary retrieval system with specific types of vocabularies and elements, and a mechanism for dependency injection.
- Refactor: Updated the `loadDataverse` function to handle specific error types and return a `TaskEither` instead of `Task`. This change improves error handling and makes the function more robust.
- Refactor: Replaced the `fetchWithAbort` function with `fetchWithSparql` and `serializeFetchResponse` functions in the `sparqlGateway` module. This change improves the handling of network requests and response serialization.
- Test: Added new test suites for the `createDepsWithOptions()` function and the process of retrieving vocabulary by type. These tests increase code coverage and ensure the correct behavior of the new functionality.
- Chore: Added a new utility function `extractErrorMessage` for extracting error messages from error objects. This function improves error handling across the application.


src/domain/vocabulary/domain.ts
```diff
@@ -0,0 +1,129 @@
...
+                    data: {
+                      ...state.data,
+                      byType: {
+                        ...state.data.byType,
+                        [type]: { ...state.data.byType[type], isLoading: true }
+                      }
+                    }
+                  }))
+                ),
+                RTE.chain(() => RTE.fromIO(() => get().data.byType)),
```
ccamel, line 76: Kleisli arrow to the rescue.

```suggestion
                RTE.chainIOK(() => () => get().data.byType),
```

src/domain/vocabulary/helper/dependencies.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+}
+
+export const withVocabularyGateway =
+  (gateway: VocabularyPort): Option =>
+  (deps: Deps) =>
+    pipe(
+      gateway,
+      E.fromPredicate((g: VocabularyPort) => R.has('retrieveVocabulary', g), GatewayTypeError),
```
ccamel, line 71: Just because an object has a property doesn't mean that the property is of the correct type (I mean signature is correct). Reflexivity (or runtime type checking) is a bad practice. The type system of Typescript is designed to catch these kinds of errors at compile time, so you don't have to perform these checks at runtime.

src/domain/vocabulary/helper/dependencies.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as P from 'fp-ts/Predicate'
+import * as N from 'fp-ts/number'
+import * as ORD from 'fp-ts/Ord'
+import * as R from 'fp-ts/Record'
+import { pipe } from 'fp-ts/function'
+import type { Show } from 'fp-ts/lib/Show'
+import type { Deps } from '../command'
+import type { VocabularyPort } from '../port'
+
+export type Option = (deps: Deps) => E.Either<VocabularyDependenciesError, Deps>
```
ccamel, line 12: It's not a `Deps`, it's a transient `Deps`, because not all the properties are set during the build.

src/domain/vocabulary/helper/dependencies.ts
```diff
@@ -0,0 +1,108 @@
...
+      }))
+    )
+
+export const createVocabularyDependenciesWithOptions = (
+  ...options: Option[]
+): E.Either<VocabularyDependenciesError, Deps> =>
+  options.reduce(
+    (acc: E.Either<VocabularyDependenciesError, Deps>, cur: Option) => pipe(acc, E.chain(cur)),
+    E.right(Object.create(null))
+  )
```
ccamel, line 108: It's not correct. As it is, I can create a dependency which is an empty object:

```ts
console.log(createVocabularyDependenciesWithOptions())
{}
```

src/domain/vocabulary/test/helper.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  language?: string
+  limit?: number
+  expectedDeps: Deps
+  expectedError?: Error
+}
+
+describe('Considering the createDepsWithOptions() function', () => {
+  describe.each`
+    vocabularyGateway                | language     | limit        | expectedDeps                                                                       | expectedError
+    ${undefined}                     | ${undefined} | ${undefined} | ${Object.create(null)}                                                             | ${undefined}
```
ccamel, line 26: According to the type, `Deps` cannot be empty. All the properties must exist and be valuated.

src/domain/vocabulary/command.ts
```diff
@@ -0,0 +1,28 @@
+import type { ReaderTaskEither } from 'fp-ts/ReaderTaskEither'
+import type { VocabularyPort } from './port'
+import type {
+  HTTPNetworkError,
+  NetworkRequestAbortedError,
+  NetworkUnspecifiedError
+} from '@/shared/error/network'
+import type { SerializationError } from '@/shared/error/serialize'
+
+export type Deps = {
```
ccamel, line 10: The type should be opaque, otherwise having a creation function doesn't provide any benefits, as it can be circumvented.

src/shared/error/serialize.ts
```diff
@@ -0,0 +1,24 @@
+import type { Show } from 'fp-ts/lib/Show'
+
+export const SerializationError = (reason: unknown) =>
```
ccamel, line 3: Why `unknown`?

src/infra/vocabulary/sparql/sparqlGateway.ts
```diff
@@ -0,0 +1,56 @@
...
+    limit: number,
+    offset: number
+  ): TE.TaskEither<RetrieveVocabularyError, RetrieveVocabularyResult> => {
+    const query = `
+      PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+	    PREFIX thesaurus: <https://ontology.okp4.space/thesaurus/>
+      SELECT ?term ?prefLabel
+      WHERE {
+        ?term a skos:Concept .
+        ?term skos:inScheme thesaurus:${type} .
```
ccamel, line 28: Injecting the type value directly into the query is a weak approach because it assumes the symbol is an accurate representation (this is a practice that many frontend developers tend to do, mainly because TypeScript encourages it). This could silently generate incorrect queries, and there's no mechanism in place for type checking or validation.

I would suggest to explicitly convert the type to its corresponding representation, I mean its URI. This is important from a functional perspective, because not all vocabularies are necessarily part of our namespace.

ccamel (changes requested):
Good job. Some comments following a quick review. The PR is substantial, and it's challenging for me to conduct a thorough review. I foresee some changes for the next steps, but we will address those when the time comes.

src/domain/vocabulary/helper/dependencies.ts
fredericvilcot: Right, so, if I understand well, there is no invariant to check for this function

src/domain/vocabulary/helper/dependencies.ts
fredericvilcot: Of course, that's a `Partial` object

src/domain/vocabulary/helper/dependencies.ts
fredericvilcot: Absolutely, I added a `validate` function that is chained to the options call that will check if all properties are declared and, if not, return a new `Either` value that might return a new Error.

src/domain/vocabulary/test/helper.spec.ts
fredericvilcot: Yep, will be modified accordingly

src/domain/vocabulary/command.ts
fredericvilcot: You're absolutely right, I have declared Deps in the `dependencies` file as an opaque type with its constructor.

src/domain/vocabulary/helper/dependencies.ts
ccamel: Yes. I don't see any preconditions to satisfy, other than those guaranteed by the typing system.

src/infra/vocabulary/sparql/sparqlGateway.ts
fredericvilcot: @ccamel 
I made the modifications in the union types, now the URI is the argument of the gateway.

src/domain/vocabulary/helper/dependencies.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export const createVocabularyDependenciesWithOptions = (
+  ...options: Option[]
+): E.Either<VocabularyDependenciesError, RetrieveVocabularyDependencies> =>
+  pipe(
+    options.reduce(
+      (acc: E.Either<VocabularyDependenciesError, TransientDeps>, cur: Option) =>
+        pipe(acc, E.chain(cur)),
+      E.right({})
+    ),
```
0xLodz, line 146: We can use the `reduce` utility from ReadonlyArray module:

```suggestion
    options,
    RA.reduce(
      E.right({}),
      (acc: E.Either<VocabularyDependenciesError, TransientDeps>, cur: Option) =>
        pipe(acc, E.chain(cur))
    ),
```

0xLodz (approved):
Amazing work! 🌶️
I Provided some minor suggestions, I let you add them if you find them useful. ☺️

src/domain/vocabulary/helper/dependencies.ts
```diff
@@ -0,0 +1,148 @@
...
+        language
+      }))
+    )
+
+export const withLimit =
+  (limit: number): Option =>
+  (deps: TransientDeps) =>
+    pipe(
+      limit,
+      E.fromPredicate(x => ORD.gt(N.Ord)(x, 0), NegativeLimitError),
```
ccamel, line 115: Not correct. When the limit is 0, it returns NegativeLimitError, but 0 isn't negative.

Instead of having specific errors for every possible domain issue, which can be many due to the different properties, I recommend using `InvalidValueError`. This error can include the property name and a reason for its rejection.

ccamel (changes requested):

src/domain/vocabulary/helper/dependencies.ts
```diff
@@ -0,0 +1,148 @@
...
+import * as S from 'fp-ts/string'
+import * as P from 'fp-ts/Predicate'
+import * as N from 'fp-ts/number'
+import * as ORD from 'fp-ts/Ord'
+import * as O from 'fp-ts/Option'
+import { pipe } from 'fp-ts/function'
+import type { Show } from 'fp-ts/lib/Show'
+import type { VocabularyPort } from '../port'
+
+const retrieveVocabularyDependenciesTag = 'retrieve-vocabulary-deps'
```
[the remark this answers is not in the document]
fredericvilcot: I wasn't aware of the [unique symbol](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#unique-symbol) notions, thanks for sharing it 🙏

src/domain/vocabulary/helper/dependencies.ts
```diff
@@ -0,0 +1,121 @@
...
+import { pipe } from 'fp-ts/function'
+import type { Show } from 'fp-ts/lib/Show'
+import type { VocabularyPort } from '../port'
+import type { Deps, RawDeps } from '../command'
+import { buildDeps } from '../command'
+
+type TransientDeps = Partial<RawDeps>
+
+type Option = (deps: TransientDeps) => E.Either<VocabularyDependenciesError, TransientDeps>
+
```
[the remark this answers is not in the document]
ccamel: It seems relevant to me. What name would you suggest?

PR #430, opened 2023-07-28
Feat/numerical input field

0xLodz:
## Numeric field
When the user is typing, the field is be formatted according to the locale, separator can be overridden using `thousandSeparator` and `decimalSeparator` props.

In order to comply with product requirement for Fees input, we can define the narrow no-break space as thousands separator and full stop as decimal separator. Also, `precision` allow to get the expected number for the KNOW token.
Comma are automatically replaced with full stop, thanks to the customizable `decimalPseudoSeparators` prop.

### Usage
```ts
<NumericField
  id={id10}
  min={0}
  onChange={handleNumericValueChange(id10)}
  placeholder="0"
  precision={APP_ENV.chains[0].feeCurrencies[0].coinDecimals}
  rightElement={<span>{APP_ENV.chains[0].currencies[0].coinDenom}</span>}
  value={singleValueField(id10)}
/>
```

### ⚠️ Important Notes
Initial Field component has been refactored with a BaseField, used for both the `TextField` and `NumericField` components. The `metadataFilling` that used Field has been updated accordingly.

Currently numeric field value is not updated properly by handlers in `metadatafilling.tsx`, because we need to format properly the stored number according to the locale, and eventually the defined separators. You can see it when you leave the field.
This issue will be handled in a dedicated PR.

### Library choice
[Maskito](https://github.com/taiga-family/maskito) 🦟 (Funny name for an input-masking library 😄 ) was chosen because it handles well caret positioning and [decimal pseudo separators](https://maskito.dev/kit/number/API). This is well maintained librairy, part of the [Taiga UI Kit](https://github.com/taiga-family/taiga-ui).


lolottetheclash (changes requested):
Thanks for your work, the numerical input component is more robust now!
however, I think we can improve a few things.
Currently, you've defined the BaseField component, which takes children.
We therefore expect to render a field, but as it's built, if the user passes another JSX element (such as an image or a paragraph), there's no problem. And in this case, the name "BaseField" no longer corresponds to what it displays.
Note also that by extending the BaseField props in the numerical input, you can pass children to this component, which is not the aim.
I was expecting the base component to already display the input and associated methods, and then the TextField and NumericalInput to add a layer of props inherent to their specific features.
I think it's possible to rework the whole thing a bit, what do you think?

0xLodz:
> Thanks for your work, the numerical input component is more robust now! however, I think we can improve a few things. Currently, you've defined the BaseField component, which takes children. We therefore expect to render a field, but as it's built, if the user passes another JSX element (such as an image or a paragraph), there's no problem. And in this case, the name "BaseField" no longer corresponds to what it displays. Note also that by extending the BaseField props in the numerical input, you can pass children to this component, which is not the aim. I was expecting the base component to already display the input and associated methods, and then the TextField and NumericalInput to add a layer of props inherent to their specific features. I think it's possible to rework the whole thing a bit, what do you think?

Thanks @lolottetheclash for your constructive and detailed feedback! ✨

The PR has been updated according to you suggestions:
- an optional render prop named `inputElement` can contain the input needed according to the context. 
- the BaseField component contains a default input html tag and associated events.
- the `metadatafilling.tsx` file as been updated with the NumericField component, and handlers has been updated.

src/ui/component/field/numericField.tsx
```diff
@@ -0,0 +1,52 @@
...
+  error,
+  leftElement,
+  rightElement,
+  ...inputProps
+}) => {
+  const { id, onChange } = inputProps
+
+  const [focused, setFocused] = useState(false)
+  const handleFocus = useCallback((): void => setFocused(true), [])
+  const handleBlur = useCallback((): void => setFocused(false), [])
```
lolottetheclash, line 23: I have a question about the way of handling blur and focus: I can see there is a classname which depends on the focus state, but in `onBlur={handleBlur}`, you don't passe the `onBlur` method given by the user.
You should add it into `handleBlur` and same for `onFocus`, what do you think?
And also, line from 21 to 23 should directly get into baseField no? same for the focused className, and it avoir you to repeat same logic for numerical and textField components.
Not sure about all this but I think it can be improved !

src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.tsx
```diff
@@ -352,10 +362,15 @@ export const MetadataFilling: FC = () => {
         value: O.none,
         render: (): JSX.Element => (
           <div className="okp4-dataverse-portal-share-data-metadata-filling fee" key={id10}>
-            <Field
+            <NumericField
+              decimalScale={APP_ENV.chains[0].feeCurrencies[0].coinDecimals} // 6
+              decimalSeparator="."
               id={id10}
-              label={'fee'}
+              lang="fr-FR"
```
lolottetheclash, line 369: Should it be better to get user current language maybe?

0xLodz: Makes sense! The field can display formatted number based on locale by default.

src/ui/component/field/numericField.tsx
0xLodz: You are definitely right, `onBlur` and `onFocus` are used here to manage field style. They don't need to be used in both Numeric and Text fields.

In order to simplify the component, onBlur and onFocus event handlers has been removed and styling has been done in CSS. Input tag has been moved as the first child of field main div to apply CSS style using `~` sibling operator, until the `:has` keyword is supported on Firefox browser.

Fields components are simplified, details of the change was done in [a862e5a](https://github.com/okp4/dataverse-portal/pull/430/commits/a862e5af1f51be331c5deeeb023dbb9572d7c71b).

lolottetheclash (approved):
Many thanks !!!

PR #431, opened 2023-07-31
Feat/floating and orientation behavior for Collapsible content

loiclaudet:
This PR allow Collapsible component to expand it's content:
- `up` or `down` (default `down`) according to the `contentExpandDirection` props.
- in a floating way (default `false`) according to the `floating` props.

Collapsible content orientation or floating behaviour can vary according to context (filers, expandable buttons, dropdown...).
This PR is a part of a refactoring of the dropdown UI PR #413. The goal is to delegate the behaviours related to the collapsible content to it's related component, `Collapsible`.

### DIFF
--- a/src/ui/component/collapsible/collapsible.scss
+++ b/src/ui/component/collapsible/collapsible.scss
@@ -1,19 +1,28 @@
 @import '@/ui/style/theme.scss';
 
+.okp4-dataverse-portal-collapsible-main {
+  display: flex;
+  flex-direction: column;
+
+  &.up {
+    flex-direction: column-reverse;
+  }
+}
+
 .okp4-dataverse-portal-collapsible-trigger-icon {
   @include with-theme {
     display: flex;
     justify-content: center;
     align-items: center;
     transition: transform 200ms ease-out;
-    
+
     &.flipped {
       transform: rotate(180deg);
     }
 
     path {
-      fill: themed("text");
-      stroke: themed("text");
+      fill: themed('text');
+      stroke: themed('text');
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ccamel (approved):

src/ui/component/collapsible/collapsible.tsx
```diff
@@ -52,17 +60,18 @@ export const Collapsible: FC<CollapsibleProps> = ({
...
           )}
         >
           <Icon name={iconName} />
         </div>
       </RCollapsible.Trigger>
       <RCollapsible.Content
-        className={classNames('okp4-dataverse-portal-collapsible-content', contentClassName)}
+        className={classNames('okp4-dataverse-portal-collapsible-content', contentClassName, {
+          floating: floatingContent,
+          [`floating-${contentExpandDirection}`]: floatingContent
```
lolottetheclash, line 73: I think here you could remove `floating: floatingContent,` and update this to :
```suggestion
          [`floating ${contentExpandDirection}`]: floatingContent
```

src/ui/component/collapsible/collapsible.scss
```diff
@@ -29,17 +38,32 @@
...
+  
+  &.floating-down {
+    top: 100%;
+  }
+
+  &.floating {
+    position: absolute;
+    z-index: 1;
+  }
+
```
lolottetheclash, line 55: to remain consistent with my remark about classnames, we could update this with:

```
  &.floating {
    position: absolute;
    z-index: 1;
    
    &.up {
      top: 0;
      transform: translateY(-100%);
    }
    
    &.down {
      top: 100%;
    }
  }
```

lolottetheclash (changes requested):
Nice work!
A remark though: I think there's a point missing: when you set the `floatingContent` prop to `true`, the content can be hidden by other elements :(

src/ui/component/collapsible/collapsible.tsx
0xLodz: That's a good trick to simplify conditional content class names. Commit has been updated according to your suggestion. Thanks!

0xLodz:
> Nice work! A remark though: I think there's a point missing: when you set the `floatingContent` prop to `true`, the content can be hidden by other elements :(

@lolottetheclash thanks for pointing this out. As discussed together and after an double check, it happens when collapsible content doesn't have a defined background color props. We would prefer let the user to customize the background according to the context.

lolottetheclash:
> > Nice work! A remark though: I think there's a point missing: when you set the `floatingContent` prop to `true`, the content can be hidden by other elements :(
> 
> @lolottetheclash thanks for pointing this out. As discussed together and after an double check, it happens when collapsible content doesn't have a defined background color props. We would prefer let the user to customize the background according to the context.

Of course, I forgot to take the background into account ^^.

0xLodz:
Floating and orientation behavior are implemented in a dedicated component in #442.

PR #436, opened 2023-08-07
fix(share): fix order in file list

AnjaMis:
This PR fixes an issue mentioned in [here](https://github.com/okp4/dataverse-portal/issues/242#issuecomment-1664135487)

### DIFF
--- a/src/ui/component/list/list.scss
+++ b/src/ui/component/list/list.scss
@@ -11,8 +11,7 @@ $transition-duration: 400ms;
     @include columns-with-gap(16px);
     list-style: none;
 
-    &.reversed-list {
-      display: flex;
+    &.reversed {
       flex-direction: column-reverse;
     }
 
--- a/src/ui/component/list/list.scss
+++ b/src/ui/component/list/list.scss
@@ -11,6 +11,11 @@ $transition-duration: 400ms;
     @include columns-with-gap(16px);
     list-style: none;
 
+    &.reversed-list {
+      display: flex;
+      flex-direction: column-reverse;
+    }
+
     .okp4-dataverse-portal-list-item-main {
       @include with-theme {
         @include noto-sans(400);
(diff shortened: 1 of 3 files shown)


lolottetheclash (changes requested):
This works perfectly!
To go a little deeper, I suggest you to create a utility function which combines the `map` and `reverse` operations and to create a test on it. 
Also, you could use `toReversed()`in order to create a new array without modifying the original one.

src/ui/page/share/dataset/steps/dataSelection/dataSelection.tsx
```diff
@@ -63,24 +63,26 @@ export const DataSelection: FC = () => {
...
+      rightElement: (
+        <div
+          className="okp4-dataverse-portal-share-dataset-page-file-deletion"
+          onClick={handleFileDeletion(id)}
+        >
+          <Icon name="bin" />
+        </div>
+      )
+    }))
+    .reverse()
```
0xLodz, line 85: If these changes are purely visual, we might prefer to change list order using CSS.
If the purpose of reversing the list is only for visual presentation, we might consider changing the order using CSS (e.g. using `flex-direction: column-reverse` for flex containers or appropriate grid properties for grid containers). This would be more efficient, especially for larger lists.

0xLodz (changes requested):

src/ui/page/share/dataset/steps/dataSelection/dataSelection.tsx
lolottetheclash: @0xLodz That's a good idea too! 
We could add an `inverted` property to the list and manage the direction according to the presence or absence of this prop.

src/ui/component/list/list.scss
```diff
@@ -11,6 +11,11 @@ $transition-duration: 400ms;
     @include columns-with-gap(16px);
     list-style: none;
 
+    &.reversed-list {
+      display: flex;
+      flex-direction: column-reverse;
```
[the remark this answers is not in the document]
AnjaMis: Ah good catch, done!

lolottetheclash (approved):
Works for me! Thanks

PR #438, opened 2023-08-10
feat: add okp4 modal

AnjaMis:
This PR adds OKP4 modal.

In addition to the modal component, this PR adds a more customized OKP4 modal component with added top and bottom section and additional appropriate styles.

## Testing

Easiest place to test the OKP4 modal component could be in the `notFoundError` page. Here is a code snippet to test:

 ``` jsx 
 const [isModalOpen, setModalOpen] = useState(false)

  const handleOpenModal = useCallback(() => {
    setModalOpen(true)
  }, [])

  const handleCloseModal = useCallback(() => {
    setModalOpen(false)
  }, [])  

```
  
and in return statement:

 ``` jsx
  <Okp4Modal
          bottomElement={<h1>This is a custom footer</h1>}
          closeOnEsc
          isCentered
          isOpen={isModalOpen}
          onClose={handleCloseModal}
          topElement={<h1>This is a custom header</h1>}
   >
    <p>
      This is the body of the modal!
    </p>
  </Okp4Modal>
```
 also don't forget to change the function in one of the page buttons:

 ```jsx

  <Button
      className="okp4-dataverse-portal-error-page-button"
      label={t('returnToHome')}
      onClick={handleOpenModal}
   />
``` 

## Screenshots

<details>

<summary> OKP4 modal light</summary> 

image

</details>

<details>

<summary> OKP4 modal dark</summary> 

image

</details>

## Notes

Additional content styles should be handled by the parent.

### DIFF
--- a/src/ui/view/modal/okp4-modal.scss
+++ b/src/ui/view/modal/okp4-modal.scss
@@ -3,39 +3,35 @@
 
 .okp4-dataverse-portal-okp4-modal-main {
   @include with-theme {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    color: themed('text');
+    padding: 12px;
     width: 358px;
+
     @include use-breakpoints(('mobile', 'tablet')) {
       width: 352px;
     }
 
-    .okp4-dataverse-portal-modal-content {
-      display: flex;
-      flex-direction: column;
-      align-items: center;
+    .okp4-dataverse-portal-okp4-modal-top-element {
+      max-height: 78px;
+    }
 
-      .okp4-dataverse-portal-okp4-modal-top-element {
-        max-height: 75px;
-        display: flex;
-        align-items: center;
-      }
+    .okp4-dataverse-portal-okp4-modal-top-element-divider {
+      border-top: 1px solid themed('modal-line-divider');
+      width: 358px;
+    }
 
-      .okp4-dataverse-portal-okp4-modal-top-element-divider {
-        border-top: 1px solid themed('modal-line-divider');
-        width: 358px;
-      }
+    .okp4-dataverse-portal-okp4-modal-content {
+      max-height: 455px;
+      overflow-y: auto;
+      padding: 0 14px 0 21px;
+      margin-top: 20px;
+    }
 
-      .okp4-dataverse-portal-okp4-modal-content {
-        height: 455px;
-        overflow-y: auto;
-        padding: 0 14px 0 21px;
-        margin-top: 20px;
-      }
-
-      .okp4-dataverse-portal-okp4-modal-bottom-element {
-        height: 75px;
-        display: flex;
-        align-items: center;
-      }
+    .okp4-dataverse-portal-okp4-modal-bottom-element {
+      max-height: 78px;
     }
   }
 }
--- a/src/ui/view/modal/okp4-modal.scss
+++ b/src/ui/view/modal/okp4-modal.scss
@@ -7,7 +7,6 @@
     flex-direction: column;
     align-items: center;
     color: themed('text');
-    padding: 12px;
     width: 358px;
 
     @include use-breakpoints(('mobile', 'tablet')) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/ui/view/modal/okp4-modal.scss [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+      width: 352px;
+    }
+
+    .okp4-dataverse-portal-okp4-modal-top-element {
+      max-height: 78px;
+    }
+
+    .okp4-dataverse-portal-okp4-modal-top-element-divider {
+      border-top: 1px solid themed('modal-line-divider');
+      width: 358px;
```
lolottetheclash, line 23: divider must take 100% modal's width according to the mockups, see here:

Capture d’écran 2023-08-11 à 14 08 54

what we have: 

Capture d’écran 2023-08-11 à 14 09 31

src/ui/view/modal/okp4-modal.scss
```diff
@@ -0,0 +1,37 @@
...
+@import '@/ui/style/mixins.scss';
+
+.okp4-dataverse-portal-okp4-modal-main {
+  @include with-theme {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    color: themed('text');
+    padding: 12px;
+    width: 358px;
```
lolottetheclash, line 11: In mockups, modal's size is set to 358px/352px (according to screens' sizes), but here your width is bigger because of the padding.

src/ui/view/modal/okp4-modal.scss
```diff
@@ -0,0 +1,37 @@
...
+
+    .okp4-dataverse-portal-okp4-modal-content {
+      max-height: 455px;
+      overflow-y: auto;
+      padding: 0 14px 0 21px;
+      margin-top: 20px;
+    }
+
+    .okp4-dataverse-portal-okp4-modal-bottom-element {
+      max-height: 78px;
```
lolottetheclash, line 34: here it should be `height`, not `max-height` as it's always the same in mockups.

src/ui/view/modal/okp4-modal.scss
```diff
@@ -0,0 +1,37 @@
...
+    .okp4-dataverse-portal-okp4-modal-top-element-divider {
+      border-top: 1px solid themed('modal-line-divider');
+      width: 358px;
+    }
+
+    .okp4-dataverse-portal-okp4-modal-content {
+      max-height: 455px;
+      overflow-y: auto;
+      padding: 0 14px 0 21px;
+      margin-top: 20px;
```
lolottetheclash, line 30: you should have 20px between bottom and content too!

src/ui/view/modal/okp4-modal.scss
```diff
@@ -0,0 +1,37 @@
...
+    color: themed('text');
+    padding: 12px;
+    width: 358px;
+
+    @include use-breakpoints(('mobile', 'tablet')) {
+      width: 352px;
+    }
+
+    .okp4-dataverse-portal-okp4-modal-top-element {
+      max-height: 78px;
```
lolottetheclash, line 18: here it should be height, not max-height as it's always the same in mockups.

src/ui/view/modal/okp4-modal.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Modal, type ModalProps } from '@/ui/component/modal/modal'
+import './okp4-modal.scss'
+
+type Okp4ModalProps = ModalProps & {
+  topElement: JSX.Element
+  bottomElement: JSX.Element
+}
+
+export const Okp4Modal: FC<Okp4ModalProps> = ({ topElement, bottomElement, ...modalProps }) => {
+  return (
```
lolottetheclash, line 11: You can remove useless return ;)

lolottetheclash (changes requested):
Nice work, just some details to fix!

src/ui/view/modal/okp4-modal.tsx
AnjaMis: Good point!

src/ui/view/modal/okp4-modal.scss
AnjaMis: yess, done!

src/ui/view/modal/okp4-modal.scss
AnjaMis: Good catch!

src/ui/view/modal/okp4-modal.scss
AnjaMis: absolutely!

src/ui/view/modal/okp4-modal.scss
AnjaMis: Ah the padding was the problem, fixed now!

0xLodz (approved):
Good job! 💪 

LGTM, even though it seems a bit restrictive to set fix heights and width.
We might prefer to set `min-height` for top and bottom elements, also `min-width` (and maybe `max-width` as well) instead of a `width` for the modal. For instance in german, we might have large title or CTA text.

We can let it as is an iterate on it with the product team if we are stuck at some point.

lolottetheclash (approved):
Thanks !

PR #444, opened 2023-08-18
Refactor/make dropdown component generic

0xLodz:
## Purpose
This PR reworks the DropDown component to make it generic and reusable. This dropDown component will be used as a base for the creation of Dropdown variants such as dropdown **with search**, **with categories** or **multiselect**.

## Usage
Example in the Home.tsx
```tsx
import { useState, type FC } from 'react'
import './home.scss'
import './i18n/index'
import { DropDown } from '@/ui/component/dropDown/dropDown'

  const [value, setValue] = useState(['1', '2', '3'])
  return (
      <DropDown
        field={<div>field, vals: {JSON.stringify(value)}</div>}
        options={
          <div>
            <button
              onClick={() => {
                setValue(s => (s.length > 0 ? [] : ['1', '2', '3']))
              }}
            >
              Toggle vals
            </button>
            <hr />
            <h2>{value}</h2>
            <p>
              lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
              incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
              lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
              incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
              lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
              incididunt ut labore et dolore magna aliqua.
            </p>
          </div>
        }
      />
    </div>
  )
}


```


src/ui/component/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,34 @@
+import type { FC } from 'react'
+import { Popover } from '@/ui/component/popover/popover'
+import './dropDown.scss'
+
+export type DropDownProps = {
+  placeholder: string
+  value: string[]
```
ccamel, line 7: Not clear to me if the componen might support multiple selections or not...

src/ui/component/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,34 @@
...
+        content={options}
+        contentClassName="okp4-dataverse-portal-dropdown-content"
+        sideOffset={8}
+        trigger={
+          <div className="okp4-dataverse-portal-dropdown-field">
+            {value.length > 0 ? (
+              field
+            ) : (
+              <p className="okp4-dataverse-portal-dropdown-field-placeholder">{placeholder}</p>
+            )}
```
ccamel, line 26: Can you clarify this? I'm having a bit of trouble understanding the purpose. Thx.

src/ui/component/dropDown/dropDown.tsx
0xLodz: Thanks for pointing this out. It serves as a base for dropdown variants, including dropdown variant with multiselect. We would prefer this component to be more agnostic, instead of coupling it with field `value`. I suggest to remove it for the base dropdown, and provide a props called for example `selections`, for the dropdown multiselect variant.
What do you think?

src/ui/component/dropDown/dropDown.tsx
0xLodz: The goal is to display a placeholder by default if no values are provided, with a style common to all dropdown variants. But after reading your previous [feedback](https://github.com/okp4/dataverse-portal/pull/444#discussion_r1301138660), using the value props might be a mistake here. We can handle the placeholder logic in the variants, that can still share a common CSS style at this component level.

AnjaMis (approved):

src/ui/component/dropDown/dropDown.tsx
```diff
@@ -0,0 +1,24 @@
+import type { FC } from 'react'
+import { Popover } from '@/ui/component/popover/popover'
+import './dropDown.scss'
+
+export type DropDownProps = {
+  field: JSX.Element
+  options: JSX.Element
```
ccamel, line 7: The name of the property is not clear to me. I'd use the names defined by the `Popover` component you use, which I think is more meaningful, e.g. `content`.

ccamel (changes requested):

src/ui/component/dropDown/dropDown.tsx
0xLodz: Renamed it using more appropriate names. Thanks for the feedback.

ccamel (approved):
Good to me! Thanks :+1:

PR #445, opened 2023-08-18
feat(metadataFilling): make know field typed as text

lolottetheclash:
This PR changes the know field from numeric type to text type.
Indeed, the `NumericField` component returns a string value.
If we had decided to store this value as a number, we would have had to transform it from string to number.
This would also have required us to convert the number to a string when retrieving the value, so as to be able to inject it back into the numeric field.
To do this, we would have had to add spaces as thousand separators and retain the ` . ` as a decimal separator, which would have meant re-encoding the logic of the `NumericField`.
Since numeric inputs retain the field value as a string, I felt it necessary to go in this direction and change `formItem` fee's type to `text`.
This allows us to keep the string as formatted, and we can always transform it into a number if required.
The numeric type has been retained in the slice, even if it's useless for the time being.

### DIFF
--- a/src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.tsx
+++ b/src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.tsx
@@ -114,14 +114,6 @@ export const MetadataFilling: FC = () => {
     [setFormItemValue]
   )
 
-  const handleNumericValueChange = useCallback(
-    (id: string) => (event: ChangeEvent<HTMLInputElement>) => {
-      const value = event.target.value
-      !isNaN(Number(value)) && setFormItemValue(id, Number(value))()
-    },
-    [setFormItemValue]
-  )
-
   const numericalFieldValue = (v: DatasetFormItem['value']) => (): string =>
     pipe(
       v,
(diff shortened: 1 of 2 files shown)


0xLodz (approved):

ccamel (commented):
I'm not entirely convinced by this approach, especially because it weakens the typing by treating a number as a string. My primary question is: why does the `NumericField` component internally store a value as a string instead of a number?

lolottetheclash:
> I'm not entirely convinced by this approach, especially because it weakens the typing by treating a number as a string. My primary question is: why does the `NumericField` component internally store a value as a string instead of a number?

@ccamel Indeed we were wrong on the type of the value of the numerical input. 
The value passed to the input must be a number, and the `NumericField's` `onChange` returns us a string (in the event). It's up to us to do the necessary transformations in order to store this value in the form of a number, and we will only have to pass it back as it is to the value of the input which should automatically format it without further manipulation.

lolottetheclash:
This PR will be closed as it is no longer relevant

lolottetheclash:
This PR will be closed as it is no longer relevant

PR #446, opened 2023-08-18
Feat/use dedicated value type for numeric and text fields

0xLodz:
Changes are related to #445 comments for numeric fields types.

## Purpose
Having distinct value type for each type of field.

## Implementation
Numeric field takes number type value as input and lift up numeric string value through the onChange callback.
The NumericField component handles internally number formatting and parsing, according to separators and locale.

## Important note
We currently fallback the numeric to `0`, but we probably want to allow the user to set back to empty value in the ShareData slice.

Instead of only updating the store if this is a valid value, we might prefer to update the initial value with `undefined`, or `NaN` if we are restricted to type number.

Then we will be able to validate (or not validate) field according to the value store.

### DIFF
--- a/src/ui/component/field/baseField.tsx
+++ b/src/ui/component/field/baseField.tsx
@@ -4,10 +4,10 @@ import classNames from 'classnames'
 import { Icon } from '@/ui/component/icon/icon'
 import './field.scss'
 
-export type BaseFieldProps = {
+export type BaseFieldProps<T> = {
   id: string
   inputElement: ReactNode
-  value?: string
+  value?: T
   label?: string
   placeholder?: string
   disabled?: boolean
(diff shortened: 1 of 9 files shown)


lolottetheclash (approved):
let's go !

src/ui/component/field/baseField.tsx
```diff
@@ -7,7 +7,7 @@ import './field.scss'
 export type BaseFieldProps = {
   id: string
   inputElement: ReactNode
-  value?: string
+  value?: string | number
```
ccamel, line 10: Why not use a bounded generic type? It feels more intuitive for specialization and inheritance to me. Just asking. You tell me.

ccamel (approved):
LGTM! Thanks. :+1:

PR #448, opened 2023-08-23
Feat/multiselect dropdown

0xLodz:
## Purpose
This PR add the **multiselect** variant of dropdown component. This is a reusable component, used for _format_ and _license_ field in the step3 of share data (see Figma's step 3 screenshot below).
image

## Testing
Update the `shareDataset.tsx` file  by only using the step 3, as follow. Then go to `http://localhost:5173/share/data`.
```tsx
import { MetadataFilling } from './steps/metadataFilling/metadataFilling'
import { Stepper } from '@/ui/component/stepper/stepper'
import type { StepElement } from '@/ui/component/stepper/stepper'
import '../i18n/index'
import './shareDataset.scss'

export const ShareDataset: FC = () => {
  const { t } = useTranslation('share')
  const metadataFilling: StepElement = { id: 'step3', content: <MetadataFilling /> }

  const steps = [metadataFilling]

  return (
    <div className="okp4-dataverse-portal-share-dataset-page-main">
      <h1>{t('share.dataset.title')}</h1>
      <Stepper steps={steps} />
    </div>
  )
}

```
## Limitation of the current implementation
When the displayed dropdown tags elements are very large, we can end with an overflow. To solve this edge case and make the component more robust, we would prefer to compute tags widths and field width in order to always display the right amount of tags, instead of relying on the `maxDisplayedTags` constant.

This edge case will be handled in a second iteration.

### DIFF
--- a/src/ui/component/tag/tag.tsx
+++ b/src/ui/component/tag/tag.tsx
@@ -8,7 +8,7 @@ import classNames from 'classnames'
 type TagProps = {
   tagName: string
   classes?: { main?: string; name?: string; icon?: string }
-  onDelete?: (tagName: string) => void
+  onDelete?: () => void
 }
 
 export const Tag: FC<TagProps> = ({ tagName, onDelete, classes }) => {
@@ -20,10 +20,10 @@ export const Tag: FC<TagProps> = ({ tagName, onDelete, classes }) => {
       e.stopPropagation()
 
       if (onDelete) {
-        onDelete(tagName)
+        onDelete()
       }
     },
-    [onDelete, tagName]
+    [onDelete]
   )
 
   return (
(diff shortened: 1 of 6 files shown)


src/ui/component/tag/tag.tsx
```diff
@@ -8,7 +8,7 @@ import classNames from 'classnames'
 type TagProps = {
   tagName: string
   classes?: { main?: string; name?: string; icon?: string }
-  onDelete?: () => void
+  onDelete?: (tagName: string) => void
```
ccamel, line 11: I don't believe it's necessary, as capturing the context can be achieved through the callback function's closure. The pattern to use is: a context is passed to the callback function only if it's essential to understand the interaction or to clarify any ambiguity from the caller perspective. It's not required in this case. Otherwise, for consistency, we would need to apply this pattern to the callbacks of all other components.

ccamel (changes requested):

src/ui/component/tag/tag.tsx
0xLodz: Oh yes! Thanks for the detailed explanation, value is already provided by the parent's context. Updated in baf4a949.

ccamel (approved):
Great! Thanks :+1:

PR #450, opened 2023-08-23
Feat/multiselect dropdown modal

0xLodz:
## Purpose
This PR adds the multiselect modal on click on `+n` tag, and display all dropdown options, as shown in the following Figma screenshot.
This modal is part of the multiselect dropdown variant component.
image

## Demo

https://github.com/okp4/dataverse-portal/assets/8344874/00afaafd-97f8-4412-af4c-ab54a3923afd

PR #451, opened 2023-08-23
feat(governance): add paragraph see more modal pop up

AnjaMis:
This PR implements a modal pop up in the governance page when the user clicks the `See more` button (when the paragraph content cannot fit in the cards).

## Screenshots

image

image

## Notes
🚨 DID detection, copy and styling will be handled in a different PR. For now the DID displaying in the pop up is not the same as in the mockups since the DID detection has not been implemented yet

### DIFF
--- a/src/ui/view/governance/details/details.scss
+++ b/src/ui/view/governance/details/details.scss
@@ -173,6 +173,27 @@
   }
 }
 
+.okp4-dataverse-portal-governance-details-modal-top-element {
+  display: flex;
+  gap: 6px;
+  align-items: center;
+
+  h1 {
+    @include norse-font(700);
+    font-size: 24px;
+  }
+}
+
+.okp4-dataverse-portal-governance-details-modal-content {
+  @include noto-sans(400);
+  max-width: 293px;
+  overflow-wrap: break-word;
+}
+
+.okp4-dataverse-portal-governance-details-modal-button {
+  width: 120px !important;
+}
+
 .okp4-dataverse-portal-governance-identity-main {
   @include with-theme {
     @include noto-sans(700);
--- a/src/ui/view/governance/details/details.scss
+++ b/src/ui/view/governance/details/details.scss
@@ -173,26 +173,31 @@
   }
 }
 
-.okp4-dataverse-portal-governance-details-modal-top-element {
-  display: flex;
-  gap: 6px;
-  align-items: center;
+.okp4-dataverse-portal-governance-details-modal-main {
+  .okp4-dataverse-portal-governance-details-modal-top-element {
+    display: flex;
+    gap: 6px;
+    align-items: center;
 
-  h1 {
-    @include norse-font(700);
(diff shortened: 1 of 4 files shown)


src/ui/view/governance/details/details.scss
```diff
@@ -173,6 +173,27 @@
   }
 }
 
+.okp4-dataverse-portal-governance-details-modal-top-element {
+  display: flex;
+  gap: 6px;
+  align-items: center;
+
+  h1 {
```
0xLodz, line 181: We might prefer to not relying on HTML tag to define style. In this case it will apply the style on every h1 from modal top element. Which is probably not what we want. Defining a class for your title is more robust here.

src/ui/view/governance/details/details.scss
```diff
@@ -173,6 +173,27 @@
...
+}
+
+.okp4-dataverse-portal-governance-details-modal-content {
+  @include noto-sans(400);
+  max-width: 293px;
+  overflow-wrap: break-word;
+}
+
+.okp4-dataverse-portal-governance-details-modal-button {
+  width: 120px !important;
```
0xLodz, line 194: Instead of relying on `!important`, you can define a custom class to the modal main element, as it was done here:
https://github.com/okp4/dataverse-portal/pull/450/files#diff-a91ad3c3a02695048c65a6b719f332328e176ab4041a4d98320b826d998c3e0cR59

But it requires [changes](https://github.com/okp4/dataverse-portal/pull/450/files#diff-6995c4f4d1c510c4a2356cd2522ecc41619af4188854fb94abe3f216b27948f7) in Okp4Modal component, which has not been merged yet!

You can probably do the same, and fix eventual conflicts if you merge it after #450.

src/ui/view/governance/details/details.tsx
```diff
@@ -13,6 +13,7 @@ import type {
 } from '@/ui/page/dataverse/zone/governance/mockedData'
 import classNames from 'classnames'
 import './details.scss'
+import { Okp4Modal } from '../../modal/okp4-modal'
```
0xLodz, line 16: We use to write full path here :)

0xLodz (changes requested):
Great job! I have provided some small request changes. Thanks

src/ui/view/governance/details/details.scss
AnjaMis: Ah yes you are absolutely right!

src/ui/view/governance/details/details.scss
AnjaMis: Good point, fixed!

src/ui/view/governance/details/details.tsx
AnjaMis: Ohhh missed that 🙈

0xLodz (approved):
LGTM!

ccamel (approved):
LGTM! :+1:

PR #452, opened 2023-08-24
Feat/add know fee wrapper in sharedata step3

0xLodz:
## Purpose
In the Step 3 of share data page, it wraps the numeric field in a component that display an associated label.
image


AnjaMis (approved):

ccamel (approved):
Good. :+1:

PR #454, opened 2023-08-29
[WIP] - Feat/create calendar component

0xLodz:
## Purpose
This PR implements a calendar base component, which is a reusable component, part of the required Datepicker component for several parts of the application (for filters on explore page, share data step 3, ...).

This component contains shared style and features for **Date selection** **Date range double view** calendar variants from the design system.

These variants will get their own reusable components, that might be used in a context of a date picker.


src/ui/component/dayPicker/dayPicker.tsx
```diff
@@ -0,0 +1,60 @@
...
+import './dayPicker.scss'
+
+const localeMap = new Map([
+  ['en', enUS],
+  ['fr', fr],
+  ['de', de]
+])
+
+type DayPickerProps = {
+  onSelect: SelectSingleEventHandler
```
ccamel, line 16: Avoid exposing the type of the handler. Additionally, if this type changes, it may break the interface of the component.

ccamel (changes requested):

src/ui/component/dayPicker/dayPicker.tsx
```diff
@@ -0,0 +1,60 @@
...
+  fromYear,
+  toYear,
+  fromDate,
+  toDate,
+  selected,
+  title,
+  isoWeek = true,
+  showOutsideDays = true
+}) => {
+  const locale = i18n.language
```
ccamel, line 38:
```suggestion
  const { i18n } = useTranslation()
  const locale = i18n.language
```

ccamel (changes requested):

src/ui/style/functions.scss
```diff
@@ -0,0 +1,19 @@
+/// Percent-encodes the "#" in a given hex color value for use in SVG or other contexts.
+/// @param {Color} $color - The hex color value to encode.
+/// @return {String} - The percent-encoded color.
+/// @example scss - Using encode-color function
+///   $encoded: encode-color(#ffffff);
+///   // $encoded: '%23ffffff'
+/// @access public
+@function encode-color($color) {
+  /// Convert the color to a string
```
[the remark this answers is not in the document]
0xLodz: Thanks for the link, will remove the comment!

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: The function signature provided by the library is the following:
```ts
(day?: Date, selectedDay: Date, activeModifiers: ActiveModifiers, e: React.MouseEvent) => void;
```
We can provide the detailed handler signature for readability and maintainability.

0xLodz: The function signature provided by the library is the following:
```ts
(day?: Date, selectedDay: Date, activeModifiers: ActiveModifiers, e: React.MouseEvent) => void;
```
We can provide the detailed handler signature for readability and maintainability.

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: You are recommending to always using the hook in the context of a component, right?

src/ui/component/dayPicker/dayPicker.tsx
ccamel: Yes. Up to you to choose what to expose, you're in control 🚗.  The idea is to design our own types rather than directly exposing types from third-party libraries. :relieved:

src/ui/component/dayPicker/dayPicker.tsx
ccamel: Yes, I think it's the best practice. Since there's the hook available, it's better than importing the `i18n` instance from the utils. That's what you've done in your PR [feat/datepicker-component](https://github.com/okp4/dataverse-portal/blob/fca8b7cba856bbafb4bdb92b81ddbc3870253c2d/src/ui/component/dateInput/dateInput.tsx#L11) and I agree with that. By the way, I'd suggest applying this rule by adding an eslint import rule...

lolottetheclash: Ooops my fault 😬 🤭

ccamel (approved):
LGTM! :+1:

lolottetheclash (approved):
Works well and looks much better without buttons, thanks!

src/ui/component/dayPicker/dayPicker.tsx
```diff
@@ -0,0 +1,67 @@
...
+const localeMap = new Map([
+  ['en', enUS],
+  ['fr', fr],
+  ['de', de]
+])
+
+type DayPickerProps = {
+  onSelect: (
+    day?: Date,
+    // @ts-expect-error built-in ract-day-picker handler signature has optional first parameter
```
fredericvilcot, line 18: Why this ts rule comment?

src/ui/component/dayPicker/dayPicker.tsx
```diff
@@ -0,0 +1,67 @@
...
+    selectedDay: Date,
+    activeModifiers: ActiveModifiers,
```
fredericvilcot, line 20: I'm not a big fan coupling the contract with the `react-day-picker` library, it violates the _dependency inversion principle_ .

src/ui/style/functions.scss
```diff
@@ -0,0 +1,16 @@
+/// Percent-encodes the "#" in a given hex color value for use in SVG or other contexts.
+/// @param {Color} $color - The hex color value to encode.
+/// @return {String} - The percent-encoded color.
+/// @example scss - Using encode-color function
+///   $encoded: encode-color(#ffffff);
+///   // $encoded: '%23ffffff'
```
fredericvilcot, line 6: I think tou can remove this line

src/ui/style/palette.scss
```diff
@@ -16,6 +17,7 @@ $secondary-disabled: #b2c3d1;
 // neutral (grey)
 $neutral: #babec2;
 $neutral-variant-1: #dfe4e9;
+$neutral-900: #4A4C4E;
```
fredericvilcot, line 20: This variables break the current palette design, even if the OKP4 design system has been updated accordingly.
I think we should rework the whole palette to take in account the changes in a very next future

fredericvilcot (changes requested):

src/ui/style/palette.scss
0xLodz: Definitely agree with your concern. @ccamel recently provided a similar [feedback](https://github.com/okp4/dataverse-portal/pull/461#discussion_r1317194833).

A [proposal](https://github.com/okp4/dataverse-portal/discussions/462) has been created to address the issue. Are you OK to let it as is and provide a solution in an dedicated PR?

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: If we don't provide error comment, TS won't compiled because of [error TS1016](https://typescript.tv/errors/#TS1016). We can find a solution to avoid it and address your [comment](https://github.com/okp4/dataverse-portal/pull/454#discussion_r1317363263) related to _dependency inversion principle_.

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: Thanks for the feedback! Didn't think about this principle. Will find a way to address it.

src/ui/style/palette.scss
fredericvilcot: I will say no, because the risk is to not handle your suggestion in the discussion, mainly because of lack of time.
And then, we will keep an incorrect and degraded version of the `palette` and, by extension, the understanding of the whole thematization..
Please, revert these variables and let's improve it in a dedicated refund PR. Thanks 🙏

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: Simply addressed because of [Function Parameter Bivariance](https://www.typescriptlang.org/docs/handbook/type-compatibility.html#function-parameter-bivariance). Didn't know that TS allows it. Since we don't need other params, we can solve it this way.

PR #457, opened 2023-09-04
Feat/improve tags field scrollbar display

0xLodz:
## Purpose
Properly display scrollbar for share data tags field.
Tags are wrapped in an intermediate html element.

image


fredericvilcot (approved):
👍

AnjaMis (approved):

PR #459, opened 2023-09-04
Feat/make popover component controlled

0xLodz:
## Purpose
Allow to control the Popover component `open` state from other triggers than the built-in `Trigger` element. It provides more flexibility, e. g. closing the popover from an element from it's content.
This update is helpful in the context of the DatePicker component.

It also remove the derived `isOpen` state from props, which is not an idiomatic React approach.

DropDown component that uses Popover has been updated accordingly.

### DIFF
--- a/src/ui/component/dropDown/dropDown.tsx
+++ b/src/ui/component/dropDown/dropDown.tsx
@@ -1,4 +1,4 @@
-import type { FC } from 'react'
+import { useCallback, type FC, useState } from 'react'
 import classNames from 'classnames'
 import { Popover } from '@/ui/component/popover/popover'
 import './dropDown.scss'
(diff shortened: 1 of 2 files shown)


src/ui/component/dropDown/dropDown.tsx
```diff
@@ -16,12 +16,20 @@ export const DropDown: FC<DropDownProps> = ({
   triggerClassName,
   contentClassName
 }) => {
+  const [open, setOpen] = useState(false)
+
+  const handleOpenChange = useCallback(() => {
+    setOpen(s => !s)
```
fredericvilcot, line 22: Please, use a more descriptive parameter name 😌

fredericvilcot (commented):
Great, just a very minor remark about consistency and readability 😌

fredericvilcot (approved):
LGTM

PR #460, opened 2023-09-04
Feat/datepicker component

0xLodz:
## Purpose
This PR implements the DatePicker component, which is the second Iteration for the creation of the **DateRangePicker** component. An overview of all steps are available in PR #454.

## Demo
https://github.com/okp4/dataverse-portal/assets/8344874/c2e434f2-882f-4be3-bb0f-ec7e04fc7c19

## Usage
Update the home.tsx file as follow:
```tsx
/* eslint-disable max-lines-per-function */
import { useState, type FC, useCallback } from 'react'
import type { DateRange } from 'react-day-picker'
import { DatePicker } from '../../component/datePicker/datePicker'
import './home.scss'
import './i18n/index'

  const [selected, setSelected] = useState<DateRange>()

  const handleSelectFrom = useCallback((date: Date | undefined) => {
    setSelected(s => ({ from: date, to: s?.to }))
  }, [])

  const year = new Date().getFullYear()

  return (
      <div style={{ display: 'flex' }}>
        <DatePicker
          fromYear={year - 100}
          onSelect={handleSelectFrom}
          selected={selected?.from}
          title="from" // TODO: i18n
          toYear={year + 20}
        />
      </div>
    </div>
  )
}

```

### DIFF
--- a/src/ui/component/dateInput/dateInput.tsx
+++ b/src/ui/component/dateInput/dateInput.tsx
@@ -2,6 +2,20 @@ import { useMemo, type FC } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Icon } from '@/ui/component/icon/icon'
 import './dateInput.scss'
+import type { PartialDateTimeFormatOptions } from '@/util/date/date'
+import { formatDateToPattern } from '@/util/date/date'
+
+const datePattern = {
+  day: 'DD',
+  month: 'MM',
+  year: 'YY'
+}
+
+const dateFormatOptions: PartialDateTimeFormatOptions = {
+  day: '2-digit',
+  month: '2-digit',
+  year: 'numeric'
+}
 
 type DateInputProps = {
   date?: Date
@@ -11,38 +25,20 @@ export const DateInput: FC<DateInputProps> = ({ date }) => {
   const { i18n } = useTranslation()
   const locale = i18n.language
 
-  const localizedDateFormatter = useMemo(
-    () => new Intl.DateTimeFormat(locale, { year: 'numeric', month: '2-digit', day: '2-digit' }),
-    [locale]
-  )
+  const dateFormatter = useMemo(() => new Intl.DateTimeFormat(locale, dateFormatOptions), [locale])
 
-  const localizedDatePattern = useMemo(
-    () =>
-      localizedDateFormatter
-        .formatToParts(date)
-        .map(({ type, value }) => {
-          switch (type) {
-            case 'day':
-              return 'DD'
-            case 'month':
-              return 'MM'
-            case 'year':
-              return 'YY'
-            default:
-              return value
-          }
-        })
-        .join(''),
-    [localizedDateFormatter, date]
+  const datePlaceholder = useMemo(
+    () => formatDateToPattern(datePattern, dateFormatOptions, locale),
+    [locale]
   )
 
   return (
     <div className="okp4-dataverse-portal-date-input-main">
       <input
         className="okp4-dataverse-portal-date-input"
-        placeholder={localizedDatePattern}
+        placeholder={datePlaceholder}
         readOnly
-        value={date ? localizedDateFormatter.format(date) : ''}
+        value={date ? dateFormatter.format(date) : ''}
       />
       <Icon name="calendar" />
     </div>
--- a/src/ui/component/dateInput/dateInput.tsx
+++ b/src/ui/component/dateInput/dateInput.tsx
@@ -1,0 +1,50 @@
+import { useMemo, type FC } from 'react'
+import { useTranslation } from 'react-i18next'
+import { Icon } from '@/ui/component/icon/icon'
+import './dateInput.scss'
+
+type DateInputProps = {
+  date?: Date
+}
+
+export const DateInput: FC<DateInputProps> = ({ date }) => {
+  const { i18n } = useTranslation()
+  const locale = i18n.language
+
+  const localizedDateFormatter = useMemo(
+    () => new Intl.DateTimeFormat(locale, { year: 'numeric', month: '2-digit', day: '2-digit' }),
+    [locale]
+  )
+
+  const localizedDatePattern = useMemo(
+    () =>
+      localizedDateFormatter
+        .formatToParts(date)
+        .map(({ type, value }) => {
+          switch (type) {
+            case 'day':
+              return 'DD'
+            case 'month':
+              return 'MM'
+            case 'year':
+              return 'YY'
+            default:
+              return value
+          }
+        })
+        .join(''),
+    [localizedDateFormatter, date]
+  )
+
+  return (
+    <div className="okp4-dataverse-portal-date-input-main">
+      <input
+        className="okp4-dataverse-portal-date-input"
+        placeholder={localizedDatePattern}
+        readOnly
+        value={date ? localizedDateFormatter.format(date) : ''}
+      />
+      <Icon name="calendar" />
+    </div>
+  )
+}
--- a/src/ui/component/datePicker/datePicker.tsx
+++ b/src/ui/component/datePicker/datePicker.tsx
@@ -1,0 +1,32 @@
+import { useState, type FC, useCallback } from 'react'
+import type { DayPickerProps } from '@/ui/component/dayPicker/dayPicker'
+import { DayPicker } from '@/ui/component/dayPicker/dayPicker'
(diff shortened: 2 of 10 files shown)


src/ui/component/dateInput/dateInput.tsx
```diff
@@ -0,0 +1,51 @@
...
+            case 'year':
+              return 'YY'
+            default:
+              return value
+          }
+        })
+        .join(''),
+    [localizedDateFormatter, date]
+  )
+
```
fredericvilcot, line 38: Utils functions should be implemented in the dedicated `util` folder and be tested accordingly

fredericvilcot (changes requested):

src/ui/component/dateInput/dateInput.tsx
0xLodz: Functions have been reworked in [7c49d90](https://github.com/okp4/dataverse-portal/pull/460/commits/7c49d909dac135b9acd3654f79891350f3fc2b1f) and tested in [43b9c1e](https://github.com/okp4/dataverse-portal/pull/460/commits/43b9c1ebee40631c343e1f6b3d3606e2491ae0b2).

Thanks for the feedback, it helped to improve function signature and behavior thanks to testing.

lolottetheclash (changes requested):
Nice job! some thoughts about utilities functions as discussed together and some few remarks. Again, nice job, date range picker is coming !

src/ui/component/datePicker/datePicker.tsx
```diff
@@ -0,0 +1,36 @@
...
+  )
+
+  return (
+    <Popover
+      align="start"
+      content={<DayPicker {...dayPickerProps} onSelect={handleSelect} />}
+      onOpenChange={handleOpenChange}
+      open={open}
+      sideOffset={8}
+      trigger={<DateInput date={selected} />}
```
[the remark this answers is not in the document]
0xLodz: We definitely can add these default props!

src/util/i18n/i18n.ts
fredericvilcot, line 1: As a global remark, I'm not sure to understand why methods related to dates are included in a `i18n.ts` file, even if they handle some locale parameters.
If we want to split `utils` functions based upon their purpose, I'll tend to create a `date.ts` file that will contain all related methods, IMHO.

src/util/i18n/i18n.ts
```diff
@@ -95,3 +95,25 @@ export const localizedNumberParser = (
 
   return (formattedString: string): number => pipe(formattedString, transformSeparators, parseFloat)
 }
+
+export type LimitedDateTimeFormatOptions = Pick<
```
fredericvilcot, line 99: Why the notion of `limited`?
You are creating your own utility function with some input type, user has no need to know if you are using, for the definition, some restricted types from another context like `Intl`.
```suggestion
export type DateTimeFormatOptions = Pick<
```

src/util/i18n/i18n.ts
```diff
@@ -95,3 +95,25 @@ export const localizedNumberParser = (
...
+  'year' | 'month' | 'day'
+>
+
+export const localizedDateFormatter = (
+  options: LimitedDateTimeFormatOptions,
+  locale: string = 'en-US'
+): Intl.DateTimeFormat => new Intl.DateTimeFormat(locale, options)
+
+export const formatDateToPattern = (
+  patternMap: Record<keyof LimitedDateTimeFormatOptions, string>,
```
fredericvilcot, line 110: I'd have named it simply `pattern` or `datePattern` because you suggest here that the input type is a `Map`, or it's a simple record.

src/ui/component/dateInput/dateInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+type DateInputProps = {
+  date?: Date
+}
+
+export const DateInput: FC<DateInputProps> = ({ date }) => {
+  const { i18n } = useTranslation()
+  const locale = i18n.language
+
+  const patternMap = useMemo(
```
fredericvilcot, line 16: As I said below, why introducing `Map` notion here?

src/util/i18n/i18n.ts
0xLodz: Yes the reason of this choice is due to the internal usage of `Intl` API. Which is probably a mistake because we already have `isoDateTime` utilities that uses `Intl` API.

What I suggest is, once updated according to your feedback, move theses functions in a dedicated `date.ts` file.

Then, in an other PR, move the functions related to `number` in a `number.ts` file.

What do you think?

src/util/i18n/i18n.ts
0xLodz: The idea was to hint that only some date format options can be provided.
The naming is probably not the most suited to express that idea.
Wanted to provide a different name that `DateTimeFormatOptions` type than `Intl` option type to avoid confusions.
May be `PartialDateTimeFormatOptions` would have been most appropriate?

Also, this function is finally just a factory that creates an Intl instance.
It might be sufficient to only use `formatDateToPattern` as utility.

What do you think?

src/util/i18n/i18n.ts
fredericvilcot: @0xLodz 
Yes, I think it's more relevant to have some utils functions based upon their purpose.
By doing this, we increase the code readability and segregation, and we will facilitate the future work on our libraries.

src/util/i18n/i18n.ts
0xLodz: You are right, I think naming I provided is terrible here.

We can either use `pattern` as you suggest, or define a `Map`, with a more appropriate name, for instance `patternByDateFormatOption`.

What would you suggest?

Thanks

src/ui/component/dateInput/dateInput.tsx
0xLodz: My bad, it was a naming mistake, some details were provided in [this answer](https://github.com/okp4/dataverse-portal/pull/460#discussion_r1321660526) from your previous feedback.

src/util/i18n/i18n.ts
fredericvilcot: @0xLodz 
You pick some properties from another type to declare your own type, IMHO, that is not limited.
It may be a matter of perspective, this is why I just added my thoughts about that.
Feel free to adjust the naming or keep like that, it must make sense for you 😌

src/util/i18n/i18n.ts
fredericvilcot: I'll suggest to keep it simple. Let's keep the `record` structure and improve the naming for its purpose 😌

lolottetheclash (approved):
Great, thanks !

src/ui/component/dateInput/dateInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }),
+    []
+  )
+
+  const dateFormatter = useMemo(
+    () => new Intl.DateTimeFormat(locale, dateFormatOptions),
+    [dateFormatOptions, locale]
+  )
+
+  const datePattern = useMemo(
```
fredericvilcot, line 39: Just as a minor remark, I would have named it `datePlaceholder` or something like that because it's not really a pattern. 

The pattern is much more the model itself in which you want to map the date. 

And, moreover, it seems a bit confusing to have the `pattern` constant in line 16 (that is a date pattern) and here another pattern variable which is the return type of the `formatDateToPattern` function.

src/ui/component/dateInput/dateInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+  )
+
+  const dateFormatOptions: PartialDateTimeFormatOptions = useMemo(
+    () => ({
+      day: '2-digit',
+      month: '2-digit',
+      year: 'numeric'
+    }),
+    []
+  )
```
fredericvilcot, line 32: Please, move these constants away from the React function to prevent from unnecessary `memo` that have a cost in memory usage.

fredericvilcot (changes requested):

src/ui/component/dateInput/dateInput.tsx
0xLodz: Oops, thanks for the feedback!

src/ui/component/dateInput/dateInput.tsx
0xLodz: yes it makes more sense, variables has been renamed. Thanks

fredericvilcot (approved):
LGTM!

PR #461, opened 2023-09-06
Feat/disabled button state

0xLodz:
## Purpose

As disabled buttons are no longer  part of the DayPicker functionality (#454), the commit for improving disabled buttons states has been cherry-picked in this PR.


src/ui/component/button/button.scss
```diff
@@ -12,13 +12,19 @@
...
   }
 
-  &:focus {
+  &:focus{
+    &.disabled {
+      outline: 1px solid themed('secondary-disabled');
+    }
+  }
+
+  &:focus:not(.disabled) {
```
lolottetheclash, line 27: is it possible to put this inside the `&:focus` ?

src/ui/style/palette.scss
```diff
@@ -15,6 +15,7 @@ $secondary-disabled: #b2c3d1;
 
 // neutral (grey)
 $neutral: #babec2;
+$neutral-200: #E5EBF0;
```
ccamel, line 18: AFAI, the naming breaks the existing convention (`-variant-x`, `-dark/light`), and the `-200` suffix doesn't clearly indicate how this color differs from the base `$neutral`.
But I don't have a clear vision of the principles governing the palette.

ccamel (changes requested):

src/ui/style/palette.scss
0xLodz: Yes, the color palettes in our design system have recently been updated, featuring clearer variations for each palette.

Here's an example of the new color palettes:
image

Refactoring the existing palettes seems essential for enhancing maintainability and readability. This can be tackled in a separate PR, once we have decided on an implementation strategy for the new palettes.

We can open a discussion to determine the most suitable solution for this matter.

As for the current PR, we can either leave it as is, provided we agree to take action on refactoring the palettes later, or directly apply the color in the specific context of the disabled button, which is currently its sole use.

ccamel: Okay, no worries. I'd suggest addressing the topic in a separate ticket (or github thread). We can consider leaving the current naming as it is. Thx.

lolottetheclash (approved):
LGTM thanks! 👌

fredericvilcot (approved):

ccamel (approved):

Issue #208, opened 2023-04-27
[Dataverse] Additional filters - Datepicker filter


ChaimaaZaoui:
# Feature: [Dataverse] Additional filters - Datepicker filter
> **As an** end-user of the portal on the `.../dataverse `page,
> **I want** to filter the dataverse by temporal coverage
> **So that** only items in a specific time period are shown

### Scenario 1 : Filter type : Datepicker 
Given that I am an end-user navigating the portal
And I am using a desktop device,
When I am on the .../dataverse page,
And I want to filter by date,
Then the following fields are displayed:

* From Datepicker
* To Datepicker

By default, no date is selected, and the datepicker displays the format DD/MM/YY.

Additionally, I expect the datepicker to have the following advanced functionalities:

* There are no select or remove buttons.
* Clicking on the datepicker opens the calendar.
* Upon clicking on a date, that date is selected, and the calendar closes.

Advanced Functionalities:
If a date is already selected:

* Clicking on a new date selects the new date and closes the calendar.
* Clicking on the same date deselects the date and automatically closes the input.

Furthermore, when I click on the datepicker, the calendar should be focused on today's date, both for the start and finish date.

The date range for the datepicker is as follows:

* Past Years: 100 years
* Future Years: 20 years

Select and Option dropdown won't be customized due to [browser limitations](https://stackoverflow.com/questions/7208786/how-to-style-the-option-of-an-html-select-element). (@0xLodz )

### Scenario 2 : Internationalisation 
ENG

> From...
> To...

>  Mo
> Tu
> We
> Th
>  Fr
>  Sa
>  Su

FR

> De...
> A...

>   Lu
>     Ma
>     Me
>     Je
>     Ve
>     Sa
>     Di

DE

> Von...
> Bis...

>Mo
>Di
>Mi
>Do
>Fr
>Sa
>So

### Mockups
Mock-ups are available on Figma Dev FIle, Page "Version 2 - FOR LATER"

![image](https://user-images.githubusercontent.com/129745194/234339703-64b1753b-6be8-4b1a-8fe3-cdf67aef4682.png)

### Environment

* staging URL: <https://app.staging.okp4.network/>

## Additional informations
None.

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206529209) by [Unito](https://www.unito.io)

ChaimaaZaoui:
DOR

- [ ] title, description, and value statement
- [ ] acceptance criteria (scenarios)
[...]

PR #463, opened 2023-09-07
Feat/dateRangePicker component

0xLodz:
## Purpose
This PR implements the DateRangePicker component, which is the last iteration to address the issue #208. The base branch corresponds to the DatePicker component PR #460.

Next step will be to integrate the DateRangePicker in ShareData step 3, and Explorer page filters.

## Usage
```tsx
import { useState, type FC, useCallback } from 'react'
import type { DateRange } from '../../component/dateRangePicker/dateRangePicker'
import { DateRangePicker } from '../../component/dateRangePicker/dateRangePicker'
import './home.scss'
import './i18n/index'

  const [selected, setSelected] = useState<DateRange>({
    from: undefined,
    to: undefined
  })

  const handleSelect = useCallback((dateRange: DateRange) => {
    setSelected(dateRange)
  }, [])

  const year = new Date().getFullYear()

  return (
      <DateRangePicker
        fromDate={selected.from}
        fromYear={year - 100}
        onSelect={handleSelect}
        selected={selected}
        toDate={selected.to}
        toYear={year + 20}
      />
    </div>
  )
}

```
## Demo

https://github.com/okp4/dataverse-portal/assets/8344874/54ba8911-bfb2-4eaa-a693-539859a12e6d


src/ui/component/dateRangePicker/dateRangePicker.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const DateRangePicker: FC<DateRangePickerProps> = ({
+  onSelect,
+  selected,
+  ...dayPickerProps
+}) => {
+  const { t } = useTranslation('common')
+
+  const handleSelectFrom = useCallback(
+    (date?: Date) => {
+      onSelect({ from: date, to: selected?.to })
```
lolottetheclash, line 22: I would have written something like:`onSelect({ ...selected, from: date})` and `onSelect({ ...selected, to: date})` but I can see it's not working because `selected` is 
```
type DateRange = {
    from: Date | undefined;
    to?: Date | undefined;
};
```

But I think we could make our own type like this:
```
type DateRange = {
    from?: Date;
    to?: Date;
};
```
I think it's more appropriate, don't you think?

lolottetheclash (changes requested):
Almost there! 
There are some discrepancies between the rendering and the mockups, I'll show you here:

Our mockup:

Capture d’écran 2023-09-11 à 08 58 07

- Start/end selected date missing style:
Capture d’écran 2023-09-11 à 08 57 55

- Dark mode start/end selected date half circle
Capture d’écran 2023-09-11 à 08 59 16
Capture d’écran 2023-09-11 à 08 59 23



- Firefox: 
Capture d’écran 2023-09-11 à 09 13 15

src/ui/component/dateRangePicker/dateRangePicker.tsx
0xLodz: Yes it would be better to not relying on `DateRange` type from the lib.
- it's easier to read in the handlers.
- we quickly get details about the type.

Thanks!

src/ui/component/datePicker/datePicker.tsx
```diff
@@ -15,8 +15,9 @@ export const DatePicker: FC<DayPickerProps> = dayPickerProps => {
   const handleSelect = useCallback(
     (date?: Date) => {
       onSelect(date)
-
-      handleOpenChange()
+      setTimeout(() => {
+        handleOpenChange()
+      }, 600)
```
fredericvilcot, line 20: Please, be careful with timer functions, you must find a way to dispose the timer at close / unmount to avoid memory leaks.

src/ui/component/dateRangePicker/dateRangePicker.tsx
```diff
@@ -0,0 +1,56 @@
+import { type FC, useCallback } from 'react'
+import { useTranslation } from 'react-i18next'
+import type { DayPickerProps } from '@/ui/component/dayPicker/dayPicker'
+import { DatePicker } from '@/ui/component/datePicker/datePicker'
+import './dateRangePicker.scss'
+
+export type DateRange = {
+  from: Date | undefined
+  to: Date | undefined
+}
```
fredericvilcot, line 10: use `?` notation instead of `| undefined`

src/ui/component/dateRangePicker/dateRangePicker.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type DateRangePickerProps = Omit<DayPickerProps, 'title' | 'onSelect' | 'selected'> & {
+  onSelect: (dateRange: DateRange) => void
+  selected: DateRange
+}
```
fredericvilcot, line 15: You couple this component with the `DayPicker` that means that it depends now on a specific contract. 
What about if we want to switch the current `DayPicker` component by another or to export it in a dedicated library ?

src/ui/component/dayPicker/dayPicker.tsx
```diff
@@ -39,20 +43,78 @@ export const DayPicker: FC<DayPickerProps> = ({
...
+    const startOfFromYear = new Date(fromYear, 0, 1)
+    const endOfToYear = new Date(toYear, 11, 31)
+
+    switch (type) {
+      case 'rangeFrom':
+        return toDate ? [{ after: toDate, before: endOfToYear }] : []
+      case 'rangeTo':
+        return fromDate ? [{ after: startOfFromYear, before: fromDate }] : []
+      case 'single':
+        return []
```
fredericvilcot, line 56: It means that, if I want to use this component as a `single` one, I can't disable any days?

What if I want, as an example, to disable in the calendar the holiday periods?

src/ui/component/dayPicker/dayPicker.tsx
```diff
@@ -39,20 +43,78 @@ export const DayPicker: FC<DayPickerProps> = ({
...
+
+  const endRange = useMemo(() => {
+    if (!fromDate || !toDate || isSameDay(fromDate, toDate)) {
+      return []
+    }
+
+    return [toDate]
+  }, [fromDate, toDate])
+
+  const selectedDayFromOtherPicker = useMemo(() => {
```
fredericvilcot, line 86: What this modifier means?

fredericvilcot (changes requested):

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: The name selectedDayFromOtherPicker is confusing because it implies a relationship with another picker component, which is not clear here. A better name would focus on the role or function that the modifier performs.

This modifier highlights the boundary date of a dateRange.
We can rename the modifier with `rangeEndpoint` for instance.

What do you think? Thanks

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: Yes, currently the component doesn't allow to disable arbitrary period, either as 'range' or 'single'.

This is a feature that we can definitely add, by providing a dedicated props.

Do we need this feature in this first version? We can easily iterate on it if needed.

src/ui/component/dateRangePicker/dateRangePicker.tsx
0xLodz: yes if we want to switch daypicker component, we should avoid to couple them.
It's probably unlikely to happen. But if it does, we can simply rewrite the Datepicker props type, without changing the component API.

The goal here was to express the relationship between both component.

If you think that we shouldn't, we can definitely update it. Please let me know! :)

src/ui/component/datePicker/datePicker.tsx
0xLodz: Sorry about that. Will update it.

src/ui/component/dayPicker/dayPicker.tsx
0xLodz: The feature has been added in [c952eda](https://github.com/okp4/dataverse-portal/pull/463/commits/c952edae323f25b7790f3132c135c4251b16239d).

cc @lolottetheclash

lolottetheclash (approved):
Nice, thanks !!! 🙏

ccamel (approved):
LGTM! :+1:

fredericvilcot (approved):

src/ui/component/dayPicker/dayPicker.tsx
fredericvilcot: Yes, sounds better, thanks

PR #469, opened 2023-09-12
Feat/date range field

0xLodz:
## Purpose
Add Temporal coverage field in share data step 3, that uses Date Range component.

## Demo

https://github.com/okp4/dataverse-portal/assets/8344874/d9c928fb-3ff7-4eb9-8650-bcde2962c931

## Summary display

image

┆Issue is synchronized with this [Wrike task](https://www.wrike.com/open.htm?id=1206524300) by [Unito](https://www.unito.io)

### DIFF
--- a/src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.scss
+++ b/src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.scss
@@ -43,7 +43,7 @@
         &.left {
           @include grid-item($colStart: 1, $colEnd: 6, $rowStart: auto, $rowEnd: auto);
           @include grid-container($colsNb: 2, $colGutter: 20px, $rowGutter: 16px);
-          grid-template-rows: 42px 42px 214px 67px 67px;
+          grid-template-rows: 42px 42px 214px 68px 68px;
 
           @include use-breakpoints(('desktop')) {
             @include grid-item($colStart: 1, $colEnd: 7, $rowStart: auto, $rowEnd: auto);
@@ -53,7 +53,7 @@
         &.right {
           @include grid-item($colStart: 6, $colEnd: 11, $rowStart: auto, $rowEnd: auto);
           @include grid-container($colsNb: 1, $colGutter: 0, $rowGutter: 16px);
-          grid-template-rows: 65px 65px 65px 205px;
+          grid-template-rows: 65px 65px 41px 136px 124px;
 
           @include use-breakpoints(('desktop')) {
             @include grid-item($colStart: 7, $colEnd: 13, $rowStart: auto, $rowEnd: auto);
@@ -64,7 +64,7 @@
           @include noto-sans(700);
           color: themed('text');
           font-size: 14px;
-          line-height: 20px;
+          line-height: 1;
           margin-bottom: 8px;
         }
       }
@@ -88,6 +88,17 @@
 
         &.fee {
           align-self: end;
+        }
+
+        .okp4-dataverse-portal-share-data-metadata-filling-temporal-coverage {
+          display: flex;
+          justify-content: space-between;
+          align-items: center;
+
+          .okp4-dataverse-portal-date-range-picker-main {
+            flex: 0 0 344px;
+            gap: 20px;
+          }
         }
       }
     }
--- a/src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.tsx
+++ b/src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.tsx
@@ -12,7 +12,8 @@ import { apply, constant, flow, pipe } from 'fp-ts/lib/function'
 import {
   type FormError,
   type InitFormPayload,
-  type DateStringRange
+  type DateStringRange,
+  isDateStringRange
 } from '@/ui/store/slice/shareData/shareData.slice'
 import { useAppStore } from '@/ui/store'
 import type { ResourceError } from '@/shared/error/resource'
@@ -175,14 +176,20 @@ export const MetadataFilling: FC = () => {
       pipe(
         id,
         formItemById,
-        IOO.map(({ value }) => value as O.Option<DateStringRange>),
+        IOO.map(({ value }) => value),
         IOO.match(
           constant({}),
           flow(
-            O.fold(constant({}), dateStringRange => ({
-              from: dateStringRange.from ? new Date(dateStringRange.from) : undefined,
-              to: dateStringRange.to ? new Date(dateStringRange.to) : undefined
-            }))
+            O.match(
+              constant({}),
+              flow(
+                O.fromPredicate(isDateStringRange),
+                O.fold(constant({}), ({ from, to }) => ({
+                  from: from ? new Date(from) : undefined,
+                  to: to ? new Date(to) : undefined
+                }))
+              )
+            )
           )
         ),
         apply(null)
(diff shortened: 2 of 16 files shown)


src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.tsx
```diff
@@ -109,6 +128,42 @@ export const MetadataFilling: FC = () => {
...
+    },
+    [setFormItemValue, localizedDateFormatter]
+  )
+
+  const dateRangeValueField = useCallback(
+    (id: string): DateRange =>
+      pipe(
+        id,
+        formItemById,
+        IOO.map(({ value }) => value as O.Option<DateStringRange>),
```
lolottetheclash, line 152: I'm not sure it's a good idea to cast it, because when we get back other form items, we always make a predicate to check that the value we are receiving is the type we are expecting. I think we should continue this way to keep robustness

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -94,6 +100,31 @@ export const Summary: FC = () => {
...
+
+        if (from && to) {
+          return (
+            <p className="okp4-dataverse-portal-share-dataset-summary-item-text">
+              {`${t('from')} ${from} ${t('to').toLowerCase()} ${to}`}
+            </p>
+          )
+        }
+
+        return (
```
lolottetheclash, line 121: there are some code duplication but I think it's more readable as you made, but I think you could maybe just refactor a bit:

```
        const dateRangeValue = getDateRangeFieldValue(item)

        if (!dateRangeValue || (!dateRangeValue.from && !dateRangeValue.to)) return null

        const { from, to } = dateRangeValue

        return from && to ? (
          <p className="okp4-dataverse-portal-share-dataset-summary-item-text">
            {`${t('from')} ${from} ${t('to').toLowerCase()} ${to}`}
          </p>
        ) : (
          <p className="okp4-dataverse-portal-share-dataset-summary-item-text">
            {from ? `${t('from')} ${from}` : `${t('to')} ${to}`}
          </p>
        )
```
We can continue to refactor to keep only one `<p>` but it's how you prefer.

lolottetheclash (changes requested):
Nice work, almost finished!

I wrote you some feedbacks ans also there are 2 things to fix:
- the one we talked together: when it's french or german, selection date is not working
- when displaying dates in `metadataFilling` step, if we change the language it has to switch to the right date display (according to the country date rules)

lolottetheclash:
@0xLodz I forgot to say that you should update the `shareDataSlice.spec.ts `file with some tests on the new `dateRange` field

src/ui/page/share/dataset/steps/summary/summary.tsx
0xLodz: Thanks for the suggestion, it's cleaner with grouped null checks!

src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.tsx
0xLodz: It has been updated in 5df3d2cc. Thanks!

src/ui/page/share/dataset/steps/metadataFilling/metadataFilling.scss
```diff
@@ -53,7 +53,7 @@
         &.right {
           @include grid-item($colStart: 6, $colEnd: 11, $rowStart: auto, $rowEnd: auto);
           @include grid-container($colsNb: 1, $colGutter: 0, $rowGutter: 16px);
-          grid-template-rows: 65px 65px 65px 205px;
+          grid-template-rows: 65px 65px 41px 120px 140px;
```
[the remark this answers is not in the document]
0xLodz: Fee field row has been set to 124px, and Tags field row size has been increased instead. Although a 16px gap is defined for the grid, last two rows have a gap larger than 16px.

lolottetheclash (approved):
Thanks for your work, let's go!

ccamel (commented):
lgtm! :+1:

ccamel (approved):

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -58,10 +62,23 @@ export const Summary: FC = () => {
...
 
+  const localizedDateFormatter = useMemo(
+    () => new Intl.DateTimeFormat(locale, { year: 'numeric', month: '2-digit', day: '2-digit' }),
+    [locale]
+  )
+
+  const formatISODate = useCallback(
+    (isoDateString: string) => localizedDateFormatter.format(new Date(isoDateString)),
+    [localizedDateFormatter]
+  )
```
fredericvilcot, line 80: These functions should be part of the `date.ts` file under utility folder and please add unitary tests.

src/ui/page/share/dataset/steps/summary/summary.tsx
```diff
@@ -94,6 +112,27 @@ export const Summary: FC = () => {
...
+            {`${t('from')} ${formatISODate(from)} ${t('to').toLowerCase()} ${formatISODate(to)}`}
+          </p>
+        ) : (
+          <p className="okp4-dataverse-portal-share-dataset-summary-item-text">
+            {from
+              ? `${t('from')} ${formatISODate(from)}`
+              : `${t('to')} ${formatISODate(to as string)}`}
+          </p>
+        )
+      }
```
fredericvilcot, line 134: A bit tricky for me to read those nested ternaries and interpolations in JSX part.
 
Could you please make it more readable 🙏?

src/ui/store/slice/shareData/test/shareDataSlice.spec.ts
```diff
@@ -122,6 +122,22 @@ describe('Given the share data slice', () => {
...
 
+  const dateStringRangeItemID6: FormItem = {
+    id: '6',
+    title: 'date range',
+    required: false,
+    type: 'date-range',
+    value: O.some({ from: '2023-09-21', to: '2023-09-30' })
+  }
+
+  const emptyDateStringRangeItemID6: FormItem = {
```
fredericvilcot, line 133: Could be good to add a case when just one property is provided and the other is `null`

fredericvilcot (changes requested):

src/ui/page/share/dataset/steps/summary/summary.tsx
0xLodz: Function has been added as utilities in [b24befa](https://github.com/okp4/dataverse-portal/pull/469/commits/b24befa5283f996c03fd5706cf65aa8458c63587), and tested in [5f001c6](https://github.com/okp4/dataverse-portal/pull/469/commits/5f001c6f45a93e9954d81999c663e58fc0259f1b). Thanks!

src/ui/page/share/dataset/steps/summary/summary.tsx
0xLodz: For sure! It has been rewrite in a more verbose, but easier to read way.

src/ui/store/slice/shareData/test/shareDataSlice.spec.ts
0xLodz: Yes. Test added! Thanks

fredericvilcot (approved):
Great, thank you

PR #475, opened 2023-09-21
Enhancement/rework palette with design system colors

0xLodz:
Rework color palette based on design system colors.

See this [discussion](https://github.com/okp4/dataverse-portal/discussions/462) for more details.


src/ui/style/palette.scss
```diff
@@ -1,50 +1,175 @@
...
+$neutral-300: rgb(216, 225, 232);
+$neutral-400: rgb(204, 215, 224);
+$neutral-500: rgb(191, 205, 217);
+$neutral-600: rgb(178, 195, 209); // Disable Outlined Button + Text on Disable Contained Button
+$neutral-700: rgb(148, 162, 174);
+$neutral-800: rgb(119, 130, 139);
+$neutral-900: rgb(89, 98, 105);
+
+// Dark blue Palette
+$darkblue-200: rgb(128, 159, 185);
```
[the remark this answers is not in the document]
0xLodz: Thanks for pointing this out!

fredericvilcot:
@0xLodz 
For my own understanding, why the major part of the `themed` functions in stylesheets have been removed in favor of static variables?

0xLodz:
> For my own understanding, why the major part of the themed functions in stylesheets have been removed in favor of static variables?

@fredericvilcot replaced themed functions in stylesheets has been removed because they were not themed related. The colors were the same, both on light and dark themes. That's why the functions were removed from theme file too, and directly used in stylesheets.

fredericvilcot:
> > For my own understanding, why the major part of the themed functions in stylesheets have been removed in favor of static variables?
> 
> @fredericvilcot replaced themed functions in stylesheets has been removed because they were not themed related. The colors were the same, both on light and dark themes. That's why the functions were removed from theme file too, and directly used in stylesheets.

Yep, that's what I understood.

But I do not agree with that, the design should be scalable and extensible.

Here, stylesheets declare by themselves hardcoded colors, with no capability to be changed without modifying the file itself and thus, with no capability to provide another theme than the provided one.

To be more compliant with the theme design, the whole app color system should rely on the theme itself, even though the color are the same (at the moment) between dark and light theme.

By doing this, you can easily swap colors from the theme file, without editing any stylesheet.

And this is important because tomorrow, the app should offer the user the capability to instantly override the global style by providing its own theme variables (e.g through a JSON file).

src/ui/style/palette.scss
```diff
@@ -1,193 +1,101 @@
...
+$darkblue-100: rgb(170, 191, 208);
+$darkblue-200: rgb(128, 159, 185);
+$darkblue-300: rgb(85, 128, 162);
+$darkblue-400: rgb(43, 96, 138);
+$darkblue-500: rgb(0, 64, 115);
+$darkblue-600: rgb(0, 53, 96);
+$darkblue-700: rgb(0, 43, 77);
+$darkblue-800: rgb(0, 32, 58);
+$darkblue-900: rgb(0, 21, 38);
+$darkblue-100: rgb(170, 191, 208);
```
Yoj16, line 51: `darkblue-100` is a duplicate ?

Yoj16 (commented):
Nice job 👏
I left you an observation 😉

src/ui/style/palette.scss
0xLodz: Good catch! duplicated variable has been removed. Thanks

Yoj16 (approved):
👏

lolottetheclash (approved):
Nice standardization work thanks !!! 👌

fredericvilcot (approved):

PR #485, opened 2023-09-26
Feat/create tooltip component

0xLodz:
## Purpose
The Tooltip component is a reusable component used in various area of the application, such as share data payment (step 5).

## Testing from home page
update the `home.tsx` file as follow:
```tsx
import { type FC } from 'react'
import { Community } from './community/community'
import { Interact } from './interact/interact'
import { Explore } from './explore/explore'
import './home.scss'
import './i18n/index'
import { Tooltip } from '@/ui/component/tooltip/tooltip'
import { Icon } from '@/ui/component/icon/icon'
import { useAppStore } from '@/ui/store'

  const theme = useAppStore(store => store.theme)

  return (
      <div className="okp4-dataverse-portal-home-page-block-container community">
        <Community />
      </div>
      <div className="okp4-dataverse-portal-home-page-block-container explore">
        <Explore />
      </div>
      <div className="okp4-dataverse-portal-home-page-block-container interact">
        <Interact />
      </div>
      <Tooltip
        content={
          <div>
            <p>
              This fee consists only of gas fees. This is an estimation based on the current network
              conditions and the complexity of the transaction. Please note: The gas fee ensures
              your transaction gets processed on the blockchain network. If the actual gas fees
              required exceed the estimated amount, the transaction will be declined. Otherwise, it
              will be validated. You will only be charged for the exact gas fees spent during the
              transaction. Additionally, the maximum processing time for a new block is 10 seconds.
              Your transaction will be included in an available block based on the gas price you
              choose. Make sure to review the gas price carefully to optimize transaction speed and
              costs.
            </p>
          </div>
        }
        contentClassName="my-custom-class-name"
        trigger={<Icon name={`info-outlined-${theme}`} />}
      />
    </div>
  )
}

```

## Overview from Figma
image

### DIFF
--- a/src/ui/component/tooltip/tooltip.scss
+++ b/src/ui/component/tooltip/tooltip.scss
@@ -1,0 +1,36 @@
+@import '@/ui/style/fonts.scss';
+@import '@/ui/style/theme.scss';
+
+.okp4-dataverse-portal-tooltip-trigger-button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 0;
+  border: 0;
+  background-color: transparent;
+}
+
+.okp4-dataverse-portal-tooltip-content {
+  @include with-theme {
+    @include noto-sans(400);
+    box-sizing: border-box;
+    max-width: var(--radix-tooltip-content-available-width);
+    max-height: var(--radix-tooltip-content-available-height);
+    padding: 10px;
+    border-radius: 6px;
+    font-size: 12px;
+    line-height: 16px;
+    background-color: themed('menu');
+    color: rgba(themed('text'), 0.6);
+    outline: none;
+    box-shadow: themed('dropdown-content-box-shadow');
+
+    .okp4-dataverse-portal-tooltip-arrow {
+      fill: themed('menu');
+
+      polygon {
+        filter: drop-shadow(0 0 2px rgba(0,0,0, 0.2))
+      }
+    }
+  }
+}
--- a/src/ui/component/tooltip/tooltip.tsx
+++ b/src/ui/component/tooltip/tooltip.tsx
@@ -1,0 +1,70 @@
+import { useEffect, useState } from 'react'
+import type { FC } from 'react'
+import classNames from 'classnames'
+import * as RTooltip from '@radix-ui/react-tooltip'
+import './tooltip.scss'
+
+type TooltipProps = {
+  content: JSX.Element
+  trigger: JSX.Element
+  contentClassName?: string
+  triggerClassName?: string
+  align?: 'start' | 'center' | 'end'
+  alignOffset?: number
+  side?: 'left' | 'right' | 'top' | 'bottom'
+  sideOffset?: number
+  portalContainer?: HTMLElement
+  delayDuration?: number
+  width?: number
+}
+
+export const Tooltip: FC<TooltipProps> = ({
+  content,
+  contentClassName,
+  trigger,
+  triggerClassName,
+  align = 'center',
+  alignOffset = 0,
+  side = 'right',
+  sideOffset = 5,
+  portalContainer,
+  delayDuration = 400,
+  width
+}) => {
+  const [containerElement, setContainerElement] = useState<HTMLElement | null>(
+    portalContainer ?? null
+  )
+
+  useEffect(() => {
+    if (containerElement) return
+
+    setContainerElement(document.getElementById('tooltip-root'))
+  }, [containerElement])
+
+  return (
+    <RTooltip.Provider delayDuration={delayDuration}>
+      <RTooltip.Root>
+        <RTooltip.Trigger
+          className={classNames('okp4-dataverse-portal-tooltip-trigger-button', triggerClassName)}
+        >
+          {trigger}
+        </RTooltip.Trigger>
+        <RTooltip.Portal container={containerElement}>
+          <RTooltip.Content
+            align={align}
+            alignOffset={alignOffset}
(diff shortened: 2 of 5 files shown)


src/ui/component/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,58 @@
...
+import './tooltip.scss'
+
+type TooltipProps = Omit<RTooltip.TooltipPortalProps, 'content'> & {
+  content: JSX.Element
+  trigger: JSX.Element
+  contentClassName?: string
+  triggerClassName?: string
+  side?: RTooltip.TooltipContentProps['side']
+  sideOffset?: RTooltip.TooltipContentProps['sideOffset']
+  container?: RTooltip.TooltipPortalProps['container']
```
fredericvilcot, line 14: Please, make props generic without letting know that, for now, we are using Radix UI.

Details of implementation should not be part of this contract.

Keep in mind that this component could be part of a future design system as it stands, using Radix UI or from scratch.

fredericvilcot (changes requested):

src/ui/component/tooltip/tooltip.tsx
0xLodz: Component has been updated according to your feedback.

We will take it into account when designing future components.

Thanks!

Yoj16:
Thanks for your testing code! 
Unfortunately, even if I set the `.my-custom-class-name {
      width: 100px;
    }` on `home.css`, the width is still the same as origin. Is it normal ? Look like we can't set the container 😊

Capture d’écran 2023-09-28 à 11 52 56

0xLodz:
> Unfortunately, even if I set the `.my-custom-class-name { width: 100px; }` on `home.css`, the width is still the same as origin. Is it normal ? Look like we can't set the container 😊

> Capture d’écran 2023-09-28 à 11 52 56

@Yoj16 thank you for your feedback!

The component updates dynamically the `width` of the content using custom properties:
image

Thus it takes precedence on the `width` defined in `my-custom-classname`:
image

However you can rely on `max-width` property to limit the tooltip to the size you want, that takes precedence on any `width` property.

Yoj16 (approved):
Nice component 🚀

src/ui/component/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+type TooltipProps = {
+  content: JSX.Element
+  trigger: JSX.Element
+  contentClassName?: string
+  triggerClassName?: string
+  side?: 'left' | 'right' | 'top' | 'bottom'
+  sideOffset?: number
+  container?: HTMLElement
+}
```
lolottetheclash, line 15: Here I would add a prop to propose to show the tooltip instantly as proposed in radix: `delayDuration` .
When I tried the tooltip I was thinking that it wasn't working because of the small delay ^^

src/ui/component/tooltip/tooltip.scss
```diff
@@ -0,0 +1,37 @@
+@import '@/ui/style/fonts.scss';
+@import '@/ui/style/mixins.scss';
```
lolottetheclash, line 2: you can remove this one

src/ui/component/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,57 @@
...
+import './tooltip.scss'
+
+type TooltipProps = {
+  content: JSX.Element
+  trigger: JSX.Element
+  contentClassName?: string
+  triggerClassName?: string
+  side?: 'left' | 'right' | 'top' | 'bottom'
+  sideOffset?: number
+  container?: HTMLElement
```
lolottetheclash, line 14: If I'm correct it's used for portals, maybe it could be expressed in the prop name to precise it

lolottetheclash (changes requested):
Nice work thanks!
just a remark: I think it's not user friendly that the users's width set in the content className isn't applied because of the radix priority. 
If you can improve this part it could be great :)

src/ui/component/tooltip/tooltip.tsx
0xLodz: Sure! Property name has been replaced with `portalContainer`.

src/ui/component/tooltip/tooltip.scss
0xLodz: Good catch, thanks!

src/ui/component/tooltip/tooltip.tsx
0xLodz: Definitely agree. `delayDuration` props has been added, with a shorter default delay (400ms).

0xLodz:
>  I think it's not user friendly that the users's width set in the content className isn't applied because of the radix priority.

@lolottetheclash thanks for your feedback.

In order to improve component usage, tooltip content CSS class [has been updated](https://github.com/okp4/dataverse-portal/pull/485/files#diff-02f8d9e218778bf4213b8ac6adf556ddd5aad7dfbbeba7106721b8134adc44fbR17), by replacing `width` by `max-width`. Thus, user can define a `width` from `contentClassName`, that overrides dynamic width provided by Radix CSS custom property, `--radix-tooltip-content-available-width`.

Also, a `width` props has been added to the component props, providing another level of customization.

Hope these changes are sufficient to provide enough flexibility and make this component easier to customize.

lolottetheclash (approved):
great, thanks !

fredericvilcot (approved):
👍

PR #487, opened 2023-09-28
Fix/popover component

0xLodz:
## Issue
The problem lies in using [refs](https://react.dev/learn/referencing-values-with-refs), which don't trigger re-renders in React. 

When `popover-root` DOM element is retrieved in `useEffect`, the component doesn't re-render, and as a result, the popover content isn't displayed.

This bug is currently masked when the Popover's parent component triggers a re-render.

## Fix
Use a `useState` hook that triggers a component re-render as soon as the `popover-root` DOM element is found.

### DIFF
--- a/src/ui/component/popover/popover.tsx
+++ b/src/ui/component/popover/popover.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from 'react'
+import { useEffect, useState } from 'react'
 import type { FC } from 'react'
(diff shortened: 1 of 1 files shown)


Yoj16 (approved):
Nice work 😉

fredericvilcot (approved):

ccamel (approved):
:+1:

PR #490, opened 2023-09-29
Feat/share data step5 tooltip

0xLodz:
## Purpose
Add tooltip info related to transaction fee.

## Demo and overview
https://www.loom.com/share/0c7cd426c66a4444ba8ce7d67012068d?sid=a241766d-8f45-44ed-8989-d6604c27f5dc


ccamel (approved):
lgtm!

Yoj16 (commented):
Good work ! 👏

Is it normal that the design is not exactly the same as the Figma ? 

The tooltip container shouldn't hide the title and the pointing arrow should be on top and not on the middle ?

### Figma design
Capture d’écran 2023-10-03 à 12 04 31

### Reality
Capture d’écran 2023-10-03 à 12 04 43

0xLodz:
> Is it normal that the design is not exactly the same as the Figma ?
> The tooltip container shouldn't hide the title and the pointing arrow should be on top and not on the middle ?

@Yoj16 good catch, thanks for your feedback! `align` and `alignOffset` props has been added to move the content at the right position. Also `arrowWidth` and `arrowHeight` props has been added to the tooltip component to fit with the design.

Yoj16 (approved):
Well done ! 👍

lolottetheclash (approved):
Works for me 👍

PR #491, opened 2023-09-29
Feat/share data payment validation summary

0xLodz:
## Purpose
Introduce a validation summary to be displayed after a successful transaction submission to the blockchain.

### Note
The actual blockchain transaction will be implemented in a future iteration. For the time being, the validation summary will appear upon clicking the "Proceed to Payment on Keplr" button.

The "Done" state for the stepper will be addressed in a separate PR.

Transaction time will be based on the real transaction once the step is wired to the blockchain. Time is always based on en-US locale.

### Demo
https://www.loom.com/share/ebcbdd509267457b9f0b9011dcd97482?sid=f507bd4c-7b13-40c6-b562-576f71c541b2


src/util/date/date.spec.ts
```diff
@@ -63,4 +68,50 @@ describe('Date utilities', () => {
       }
     )
   })
+
+  describe('Considering the formatISODateToParts function', () => {
```
lolottetheclash, line 72: As discussed, I think that this test is useless as I have the feeling it is only testing the API, not a custom function. @fredericvilcot am I missing something?

package.json
```diff
@@ -18,8 +18,8 @@
     "dev": "vite --host",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "test": "jest",
-    "test:watch": "jest --watch --no-cache",
+    "test": "TZ=UTC jest",
+    "test:watch": "TZ=UTC jest --watch --no-cache",
```
lolottetheclash, line 22: I think this configuration may cause some problems, so I would recommand to set the time zone directly in the test to improve its determinism.

lolottetheclash (changes requested):
Looking nice, as always :)
I let you check the remarks below to improve just a little bit

package.json
0xLodz: Definitely agree to limit timezone config to the test.

It has been defined [directly in the test](https://github.com/okp4/dataverse-portal/compare/b4eeaa4f10dad3c6d2718743dd8fb5c1891c7ee4..dee74bba8e639ac7d0e18bbe212034156f4dd7ef#diff-e605af1e3f46144bd488dd4893377c8f45ca2afd8f43ddd14011db3c288ea8e8R81) according to your feedback.

Thanks!

Yoj16 (approved):
Well done ! 😊

lolottetheclash (approved):
Nice, thanks 🚀

